PR #42, opened 2018-01-30
Vehicle similarity search module (ETA components)

chrisstauffer:
common core components to first pass encoding and matching capabilities.


jasoncorso:
Is this now to be reviewed?

eta/core/features.py
```diff
@@ -25,118 +25,168 @@
...
+
+        Args:
+            data: the data to featurize
+
+        Returns:
+            the feature vector
+        '''
         self.start(warn_on_restart=False, keep_alive=False)
-        v = self._featurize(data)
+        fv = self._featurize(data)
```
jasoncorso: 2-char variable names are just as bad as 1-char variable names!  And I am fine with either in cases like this.

eta/core/numutils.py
```diff
@@ -71,12 +71,21 @@ class GrowableArray(object):
...
+        Args:
+            rowlen: the desired length of each row
+        '''
         self.data = []
         self.rowlen = rowlen
 
     def update(self, row):
         '''Add row to array.'''
-        assert len(row) == self.rowlen, "Expected row length %d" % self.rowlen
+        if len(row) != self.rowlen:
```
jasoncorso: Unclear to me what the difference is in functionality or error-checking.

eta/core/types.py
```diff
@@ -676,3 +718,195 @@ class FrameSequence(JSONFileSequence):
...
+    '''A sequence of features indexed by one numeric parameter.
+
+    Examples:
+        /path/to/features/%05d.npz
+    '''
+
+    @staticmethod
+    def gen_path(basedir, params):
+        return os.path.join(
+            basedir, "{name}", "{idx}.npz").format(**params)
```
jasoncorso: issue should be created to allow generalization beyond just npz here and elsewhere.

eta/core/numutils.py
brimoor: just following the developing convention to raise custom exceptions that categorize our errors

eta/core/types.py
brimoor: Yeah and here we'd create sub-types of `Features` to encapsulate the various implementations

PR #71, opened 2018-06-14
Integrate primitive operations to eta/core

kunyilu:
1. Integrate Dense Optical Flow into ETA
    - Farneback's method
2. Integrate online background subtraction into ETA
    - MOG2 background subtractor
    - KNN background subtractor
3. Integrate edge detection into ETA
    - Canny edge detection
4. Integrate common feature point detection algorithms into ETA
    - Harris corners
    - FAST feature points
    - ORB feature points


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2017, Voxel51, LLC
+Copyright (c) 2017, KUNYI
```
brimoor: lol! you decided to claim copyrights for the entire ETA library now?

Please revert this

kunyilu: Sorry for that. I did not realize it. solved

eta/core/primitives.py
```diff
@@ -0,0 +1,605 @@
...
+        self.poly_n = poly_n
+        self.poly_sigma = poly_sigma
+        self.use_gaussian_filter = use_gaussian_filter
+
+        self._prev_frame = None
+        self._flags = (
+            cv2.OPTFLOW_FARNEBACK_GAUSSIAN if use_gaussian_filter else 0)
+
+    def _process_frame(self, img):
+        curr_frame = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
```
jasoncorso: ACK.  We recently decided to establish and enforce a contract that all images ETA reads will be converted to RGB from BGR (when ingested as such).  This call violates that contract.  (Of course, this is color-to-gray, so it is not fatal, but nevertheless should be fixed.)

eta/core/primitives.py
```diff
@@ -0,0 +1,605 @@
...
+                # Compute edges
+                edges = self._process_frame(img)
+
+                if masks_path:
+                    # Write edges mask
+                    np.save(masks_path % p.frame_number, edges.astype(np.bool))
+
+                if vid_path:
+                    # Write edges video
+                    p.write(cv2.cvtColor(edges, cv2.COLOR_GRAY2BGR))
```
jasoncorso: Again color contract issue.

eta/core/primitives.py
```diff
@@ -0,0 +1,605 @@
...
+            k (0.04): Harris detector free parameter
+        '''
+        self.threshold = threshold
+        self.block_size = block_size
+        self.aperture_size = aperture_size
+        self.k = k
+
+    def _process_frame(self, img):
+        # works in OpenCV 3 and OpenCV 2
+        gray = cv2.cvtColor(img, cv2.COLOR_BGR2GRAY)
```
jasoncorso: Again COLOR.

eta/core/video.py
```diff
@@ -27,15 +27,109 @@
 import cv2
 import numpy as np
 
-import eta.core.image as im
 from eta.core import utils
+import eta.core.image as im
```
jasoncorso: This does not follow the convention
`from eta.core import utils as etau` or just `import eta.core.utils at etau` and
`import eta.core.image as etai`

eta/core/video.py
```diff
@@ -195,6 +292,12 @@ def __init__(
                 each frame range when the write() method is called. When
                 out_vidpath is None or "", no videos are written
 
+            out_single_vidpath: a path like "/path/to/video.mp4" that specifies
```
jasoncorso: I think this should have been added in a different PR; this is rather peculiar functionality and probably deserves more discussion.  Alas, time is not too abundant to do that.

eta/core/primitives.py
brimoor: I will do this in a separate PR. As currently written, ETA loads images in BGR format

eta/core/video.py
brimoor: This cleanup is coming in the `veh-sim` branch.

eta/core/video.py
brimoor: It is awkward. But sometimes we want to process sub-clips, in which case we need `"/path/to/video/%05d-%05d.mp4"`, and sometimes we want to process the entire video, in which case we just want to say `"/path/to/video.mp4"`...

The better practice is probably to introduce `VideoClipsProcessor` to handle clips, but alas this works and is backwards compatible for now.

PR #72, opened 2018-06-15
Adding a tool to automatically generate module metadata files from source

brimoor:
This pull request introduces an `eta/scripts/generate_module_metadata.py` script that can generate metadata JSON files for ETA modules programmatically from the (docstrings in the) `.py` file.

The syntax is:
```shell
python generate_module_metadata.py /path/to/eta_module.py
```

See the updated `docs/modules_dev_guide.md` or the modules in `eta/modules` for examples of the required docstring format.

The script was used to successfully regenerate the module metadata files for all modules in `eta/modules`


jasoncorso:
Will review, but want to make a comment first that I think the usage might rather (also?) be exposed through the common `eta` "binary" interface, such as `eta --metadata foo.py`.

eta/core/module.py
```diff
@@ -404,19 +476,20 @@ class ModuleMetadata(Configurable, HasBlockDiagram):
     '''Class the encapsulates the architecture of a module.
```
jasoncorso: Noticed this typo when reviewing this new code.  `Class that` not `Class the`

eta/core/pipeline.py
```diff
@@ -330,40 +330,40 @@ class PipelineInput(object):
...
 
 class PipelineOutput(object):
     '''Class describing a pipeline output.
 
     Attributes:
         name: the output name
-        node: the ModuleNode instance of the module output that the pipeline
-            output is connected to
+        output: the ModuleOutput instance of the module output that the
+            pipeline output is connected to
```
jasoncorso: From this design, it seems a Module can have only one output?  That is not what I recall observing in the various block diagrams recently.  What am I missing?

eta/modules/embed_vgg16.py
```diff
@@ -49,15 +57,28 @@ def __init__(self, d):
...
+
+    Outputs:
+        backing_path (eta.core.types.Directory): The directory in which to
+            write the embeddings
+    '''
 
     def __init__(self, d):
-        self.backing_path = self.parse_string(d, "backing_path")
         self.video_path = self.parse_string(d, "video_path")
+        self.backing_path = self.parse_string(d, "backing_path")
```
jasoncorso: Confused why these are swapped...

eta/core/module.py
brimoor: thanks- fixed.

eta/core/pipeline.py
brimoor: Modules and pipelines can have multiple outputs. What this says is that a pipeline output must be connected to exactly one module output (unlike pipeline inputs, which may feed their values to the inputs of multiple modules)

eta/modules/embed_vgg16.py
brimoor: `video_path` is the input and `backing_path` is the output, so my eyes want to see the input listed first

brimoor:
Thanks for the suggestion to support this through the `eta` command line tool. Not sure why I keep forgetting about that! I'll do it

eta/modules/embed_vgg16.py
jasoncorso: Ah, I see.  They were originally alpha-sorted.  I suppose this is fine.

PR #77, opened 2018-06-17
Improving handling of default values for parameters

brimoor:
This pull request simplifies the definition of module parameters for metadata purposes. Now module parameters can be either be:
- `required`: the user must provide a value
- `optional`: the user may provide a value for the parameter, but a default value is provided in case they don't
(Previously there was a confusing hybrid notion of a required parameter that provides a default value)

This PR also:
- updates the module metadata generation tool to require default values to be expressed in Python format, not JSON. This is more intuitive as the values themselves are parsed from `.py` files
- fixes a bug in the module metadata generation script that caused stray `default` attributes to appear in the definitions of required fields

### DIFF
--- a/eta/core/module.py
+++ b/eta/core/module.py
@@ -272,10 +272,19 @@ class ModuleParameterConfig(Config):
         self.type = self.parse_string(d, "type")
         self.description = self.parse_string(d, "description")
         self.required = self.parse_bool(d, "required", default=True)
-        self.default = self.parse_raw(d, "default", default=None)
+        if not self.required:
+            self.default = self.parse_raw(d, "default")
+        elif "default" in d:
+            raise ConfigError(
+                "Module parameter '%s' is required, so it should not have a "
+                "default value" % self.name)
 
     def attributes(self):
-        return ["name", "type", "description", "required", "default"]
+        attrs = ["name", "type", "description", "required"]
+        if not self.required:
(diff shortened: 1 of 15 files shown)


eta/core/module.py
```diff
@@ -452,24 +453,37 @@ def is_data(self):
...
+            raise ModuleMetadataError(
                 "Module parameter '%s' has type '%s' but must be a subclass "
-                "of Builtin or ConcreteData") % (name, type_))
+                "of Builtin or ConcreteData" % (name, type_))
         return type_
 
     def _validate_default(self):
-        if self.is_builtin:
-            is_valid = self.type.is_valid_value(self.default)
+        if self._default is None:
```
jasoncorso: Only question regards the use of `no_default`.  If a parameter is `required and has `no_default`, then how does this bubble down to this `_validate_default` function?  It would get to the last `if not is_valid` as a `True`, but should we output a more descriptive note for this case, which is a bit different than the case when a developer or user sets it to an invalid but not the `no_default`.

brimoor: This function is only called for optional parameters, since required parameters don't have defaults

brimoor:
@jasoncorso I cherry-picked the commits that fix the pipeline-with-a-single-module bug from `veh-sim` into this branch.

PR #80, opened 2018-06-19
Exposing per-frame methods of primitives

brimoor:


eta/core/primitives.py
```diff
@@ -148,7 +164,7 @@ def __init__(
         self._flags = (
             cv2.OPTFLOW_FARNEBACK_GAUSSIAN if use_gaussian_filter else 0)
 
-    def _process_frame(self, img):
+    def process_frame(self, img):
```
kunyilu: @brimoor since this function only return the output in cartesian format, is your idea to transform the output to the other two formats(polar and visualization image) in the rapid motion detection module.

eta/core/primitives.py
```diff
@@ -66,35 +66,51 @@ def process(
...
+        raise NotImplementedError("subclass must implement process_frame()")
 
-    def _reset(self):
+    def reset(self):
         '''Prepares the object to start processing a new video.'''
         pass
 
+    @staticmethod
+    def cart_to_polar(flow_cart):
+        '''Converts the Cartesian optical flow vectors to polar form.
```
kunyilu: Brian, I wonder why you make cart_to_polar a static method of the class this time? Any purpose?

brimoor: I made it a static method of the optical flow class so its clear that this functionality is related specifically to optical flow. However, upon second thought, I should probably just make it a top-level method so the user doesn't have to write `primitives.DenseOpticalFlow.cart_to_polar(...)`, which is a mouthfull.

eta/core/primitives.py
brimoor: I refactored this functionality into a standalone method so that the user of this class can convert their Cartesian flow fields to polar if they want to.

The reason I don't return all formats (cartesian, polar, and HSV image) by default is that the caller might not want all of these formats, so we should avoid doing any unnecessary computation. We should be especially careful about avoiding unnecessary computation in `core` modules

PR #83, opened 2018-06-22
Adding a model management system to ETA

brimoor:
This pull request introduces a cloud-backed model management system for ETA.

The framework allows models to be referenced (loaded, downloaded, etc.) by a simple names like `my-model`. It also supports versioned models with the syntax `my-model@1.0`. See `docs/models_dev_guide.md` for more details.

ETA models are stored in Google Drive with publicly accessible links, and the system can be easily extended to support different model storage mechanisms

This PR also updates the existing VGG-16-related code to use the new models classes.


docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+to be archived and backed up to ensure that they remain secure and readily
+accessible.
+
+The ETA library provides a simple, yet powerful interface that satisifies
+these constraints.
+
+
+## Design
+
+The `eta.core.models` module implements the core interface for managing models
```
jasoncorso: Wondering if we want to be clear with our usage of "module" in ETA documentation.  This here is a "python module".

docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+in ETA. The underlying design principle of the interface is that all models
+are stored in cloud storage and can be easily downloaded to a local ETA
+installation as needed using a simple version-aware naming system.
+
+### Model versions
+
+Each model is identified by a unique string of the form `<name>@<version>`,
```
jasoncorso: How is uniqueness of strings enforced across developers?

docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+explicitly using the `<name>@<version>` syntax. The `<name>` syntax can
+also be used with versioned models, in which case the latest version of the
+model (i.e. the largest version number) will be referenced.
+
+#### Models search path
+
+The `models_dirs` field of the ETA config defines the list of directories to
+use when performing all models-related operations. If you add a new models
+directory to your path, you can use the `eta.core.models.init_models_dir`
+function to initialize it for use with the model management interface.
```
jasoncorso: This here probably needs a bit more explanation.  When would you need to?  I am guessing only if it is added mid-execution, otherwise it will get loaded/initialized during startup?

docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+The following is an example contents of a `manifest.json` file:
+
+```json
+{
+    "models": [
+        {
+            "base_name": "VGG-16",
+            "base_filename": "vgg16_weights.npz",
+            "version": null,
+            "manager": {
```
jasoncorso: It's fine to stick with this, but I think it is awkward to reference  "manager" field inside of the "models" description.  And, especially including the model file id in the manager field.  Confused why it is organized like this.

jasoncorso: I suppose you answer this below.  Still noting a bit of confusion, but I get it.

docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+folder. The `eta.core.models.ETAModelManager` class provides the underlying
+functionality used to download these models.
+
+> Note: For security reasons, the ETA model manager is unable to upload and
+> delete models. These tasks must be performed manually by a Voxel51
+> administrator.
+
+#### Model weights
+
+The `eta.core.models.ModelWeights` class defines the base implementation of
```
jasoncorso: Should we note that "weights" is ETA's term for any and all parameters or model coefficients?  EG Linear SVM coefficients would also be stored in these weights.

docs/models_dev_guide.md
```diff
@@ -0,0 +1,263 @@
...
+The ETA config provides a `max_model_versions_to_keep` attribute that allows
+you to configure the maximum number of versions of a given model to keep on
+disk. If this limit is execeeded, the oldest versions of a model are deleted
+until the limit is satisifed again.
+
+The models framework automatically checks this condition and performs the
+necessary cleanup after each time a model is downloaded. However, you can
+manually invoke the cleanup at any time by running:
+
+```py
```
jasoncorso: Maybe this should be performed during initialization or cleanup of ETA?  It seems risky to require an interactive function call to manage model lifecycle.  (I'm envisioning long-running installations...)

eta/core/models.py
```diff
@@ -0,0 +1,909 @@
...
+    path, so models in this directory will not be findable until you update
+    your ETA config.
+
+    Args:
+        new_models_dir: the directory to initialize
+
+    Raises:
+        ModelError: if the models directory is already initialized
+    '''
+    if ModelsManifest.dir_has_manifest(new_models_dir):
```
jasoncorso: To not require a user to manually initialize the directory s/he has added to her/his config, we could add a bit of code to the eta initialization that call this for each models directory and just catches this Exception and moves on.

eta/core/models.py
```diff
@@ -0,0 +1,909 @@
...
+            the model weights
+        '''
+        download_model(self.model_name, force=False)
+        return self._load()
+
+    def _load(self):
+        raise NotImplementedError("subclass must implement _load()")
+
+
+class NpzModelWeights(ModelWeights, dict):
```
jasoncorso: I get that these are to be viewed as ReadOnly but where/how are suitable files created?  By a different class/interface?  Any such model suffices?  This seems like one-half of the solution.

eta/core/vgg16.py
```diff
@@ -44,28 +44,19 @@
...
     '''Configuration settings for the VGG16 network.'''
 
     def __init__(self, d):
-        self.weights = self.parse_object(d, "weights", WeightsConfig)
-
-    @classmethod
-    def load_default(cls):
-        '''Loads the default config file.'''
-        return cls.from_json(DEFAULT_CONFIG_PATH)
+        self.model = self.parse_string(d, "model", default="VGG-16")
```
jasoncorso: Heavens why add the dash when everywhere else it is VGG16!!

eta/core/vgg16.py
```diff
@@ -434,33 +425,27 @@ def _build_fc_layers(self):
...
+class VGG16FeaturizerConfig(VGG16Config):
     '''Configuration settings for a VGG16Featurizer that works on images.'''
-
-    def __init__(self, d):
-        self.weights = self.parse_object(
-            d, "weights", WeightsConfig, default=None)
-        if self.weights is None:
-            self.default_config = VGG16Config.load_default()
-            self.weights = self.default_config.weights
+    pass
```
jasoncorso: I suppose it is fine to keep this, but seems pointless now.  The earlier implementation let the user specify a different weights for a featurizer, but I don't think that functionality was ever used.

docs/models_dev_guide.md
brimoor: I think its clear in this case because only Python modules use the `eta.core.XXX` syntax

docs/models_dev_guide.md
brimoor: Uniqueness is forced via the `manifest.json` files on your models search path. Basically any models-related operation you perform will cause all manifests to be read. If there is a name conflict at load time, an error will be raised.

Note that we need to commit the manifest files to the repo, because this is how we "expose" our models across the team.

We can also have our own "private" models for each developer by simply adding our own private folders to our search paths and not adding the relevant `manifest.json` to version control.

docs/models_dev_guide.md
brimoor: `manager` is written like this because it is a `Configurable` class like `FeaturizerConfig` that can construct subclasses of itself on-demand. This allows us to support remote remote storage solutions.

ETA models need to be publicly accessible, but THETA models need to be private, so the information in `manager` depends on the cloud storage strategy. Everything else is local information needed to work with the models on-premises.

docs/models_dev_guide.md
brimoor: Yeah good point- on startup is definitely preferable to mid-download. I'll update

eta/core/models.py
brimoor: What does suitable files mean? When one calls the `load` method of a `ModelWeights` class, it looks up the model `name` in the models manifest, automatically downloads it if necessary, loads the weights, and returns them to the user.

What's missing?

jasoncorso: I just mean the actual model files.  My read of the structure here is that this code awesome there exists other code for creating models and their files; then it is up to the developer to create this manifest information.  ...just seems like there is a disconnect here that the same class is not being used during the model creation as this ultimate model management/storage/etc.  (Or mixin.)  Something to think about; not critical.

eta/core/vgg16.py
brimoor: We can update when a convention arises

eta/core/vgg16.py
brimoor: It's still possible to specify different weights

docs/models_dev_guide.md
brimoor: Good point. I'll clarify.

eta/core/models.py
brimoor: New models are created using `theta.core.models.publish_model`. It's a simple one-liner that sets everything up- uploads model to the cloud, logs it in a manifest, etc.

I had to split it up because there are certain proprietary aspects of our codebase that we don't want to put in ETA. We expose models to the world in ETA through a public Google Drive folder, but we can't let other people publish models to that location. And our code `theta.core.storage` code for reading/writing files to private Google Drive folders is not in ETA, so the aspects related to "publishing" THETA modules had to be split off into THETA.

docs/models_dev_guide.md
brimoor: Done

docs/models_dev_guide.md
brimoor: Done

PR #90, opened 2018-06-25
add some new data types for tracking project

kunyilu:
Added some new data types for the purpose of specifying inputs/outputs in metadata files.

### DIFF
--- a/eta/core/types.py
+++ b/eta/core/types.py
@@ -676,3 +676,72 @@ class FrameSequence(JSONFileSequence):
     '''
 
     pass
+
+
+class EmbeddedObjectsSequence(FrameSequence):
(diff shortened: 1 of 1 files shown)


brimoor:
@kunyilu I renamed some of the types you created, because your tracking modules are actually writing `EmbeddedFrame` data classes, not `EmbeddedObjects` (same for other classes).

Also note that I added the non-sequence versions of the various types you added. If one adds a "Sequence" type, it is good manners to also add a type for the non-sequence version, even though you might not explicitly use it.

Along similar lines, you didn't add types for some of the other tracking data classes. If they might be useful to read by other modules, you should also declare types for them, too.

PR #93, opened 2018-06-26
C3D featurizer

jinyixin621:
Integrated an open source implementation of C3D into ETA.  Ability to use pretrained weights to embed a fixed length and spatial resolution video clip into the C3D space.
It can input a video clip, parameters about how to sample it temporally (take the first K frames, slide a K-frame window, or resize to just K frames), and output the embedding.


eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class C3DConfig(Config):
+    '''Configuration settings for the C3D network.'''
+
+    def __init__(self, d):
+         self.model = self.parse_string(d, "model", default="C3D")
```
brimoor: This is never used. I suspect you copied this from VGG-16 without understanding what it does. Don't copy code you don't understand!

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+         self.dropout = self.parse_number(d, "dropout", default="0.6")
+         self.batchsize = self.parse_number(d, "batchsize", default="1")
+         self.model_path = self.parse_string(d, "model_path", default="../eta/models/sports1m_finetuning_ucf101.model")
```
brimoor: reminder: style guide says 79 characters or less per line. Please configure your editor to show a vertical ruler so you see when you need to do a line break.

Ask Kunyi for help if needed.

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class C3DConfig(Config):
+    '''Configuration settings for the C3D network.'''
+
+    def __init__(self, d):
+         self.model = self.parse_string(d, "model", default="C3D")
+         self.dropout = self.parse_number(d, "dropout", default="0.6")
```
brimoor: you declare these fields as numbers, but you provide default values that are strings. The default value here should be `0.6`, not `"0.6"`

eta/core/c3d.py
brimoor: With that said, please do keep this field here, because you need to update the way the model is loaded here. While you were sick we introduced a model registry that allows us to "publish" models and refer to them by name.

Please refer to `vgg16.py` to see how this works. Basically all models will be storead Google Drive and your `_load_model` code here should use the `eta.core.models` tools to download the model if necessary and get the path to the model without you having to hard code it in `model_path` below.

Ask me for help on how to actually "publish the model".

eta/core/c3d.py
brimoor: This comment applies throughout.

eta/core/c3d.py
brimoor: See my above comment on `model`. This field should be removed.

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+        Args:
+            inpath: path to the input video
+            k: number of frames to extract for the input video
+            out_size: size of the output frames
+        '''
+        data = []
+        num_frames = etav.get_frame_count(inpath)
+        assert k < num_frames
+        with etav.VideoProcessor(inpath, "1-%d" % k, out_size=out_size) as p:
```
brimoor: `VideoProcessor` is for reading a video and then writing a new video in response. You should use `FFmpegVideoReader` here because you're just reading a video

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+        self.sample_method = self.config.sample_method
+        self.c3d  = None
+
+    def get_first_k_frames(self, inpath, k, out_size):
+        '''Sample first k frames in a video.
+
+        Args:
+            inpath: path to the input video
+            k: number of frames to extract for the input video
+            out_size: size of the output frames
```
brimoor: `out_size` is a misleading name. Nothing is output, this is the dimension to use when embedding the frames

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+                'bd1': self._variable_with_weight_decay('bd1', [4096], 0.04, 0.0),
+                'bd2': self._variable_with_weight_decay('bd2', [4096], 0.04, 0.0),
+                'out': self._variable_with_weight_decay('bout', [101], 0.04, 0.0),
+                }
+        self.build_c3d(self.imgs, self.config.dropout, self.config.batchsize,
+            weights, biases)
+        self._load_model()
+
+    def _variable_on_cpu(self, name, shape, initializer):
+        with tf.device('/cpu:0'):
```
brimoor: You are forcing CPU 0 here, but we want to use GPU! And we shouldn't be forcing any specific devices.

See `vgg16.py`. No devices are forced.

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+                feed into the network
+            layer: an optional layer whose output to return. By default, the
+                output softmax layer (i.e., the class probabilities) is
+                returned
+        '''
+        if layer is None:
+            layer = self.out
+
+        return self.sess.run(layer, feed_dict={self.imgs: [imgs]})[0]
+    def close(self):
```
brimoor: one space between methods in a class

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+        return var
+
+    def _variable_with_weight_decay(self, name, shape, stddev, wd):
+        var = self._variable_on_cpu(name, shape, tf.truncated_normal_initializer(stddev=stddev))
+        if wd is not None:
+            weight_decay = tf.nn.l2_loss(var) * wd
+            tf.add_to_collection('losses', weight_decay)
+        return var
+
+    def conv3d(self, name, l_input, w, b):
```
brimoor: these should be private methods (`_conv3d`, `_max_pool`, `_build_c3d`)

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+class C3DFeaturizer(Featurizer):
+    def __init__(self, config=None):
+        super(C3DFeaturizer,self).__init__()
+        self.config = config or C3DFeaturizerConfig.default()
+        self.validate(self.config)
+        self.sample_method = self.config.sample_method
+        self.c3d  = None
+
+    def get_first_k_frames(self, inpath, k, out_size):
```
brimoor: These `get_` methods are more general than just `C3D`. They just extract frames from a video. They should be upgraded top-level functions in this module.

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+            rng.append(num_frames - 1)
+        elif len(rng) > k:
+            rng.pop()
+        assert len(rng) == k
+        with etav.VideoProcessor(inpath, out_size=out_size) as p:
+            for counter, img in enumerate(p):
+                if counter in rng:
+                    img = etai.resize(img, out_size, out_size)
+                    data.append(img)
+        np_arr_data = np.array(data).astype(np.float32)
```
brimoor: This returns a tensor whose first dimension is number of frames. You should follow the typical convention of `m x n x k`, where `k` = number of frames

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+                scalling the close() method of this class when you are done
+                computing
+            imgs: an optional tf.placeholder of size [XXXX, 16, 112, 112, 3] to
+                use. By default, a placeholder of size [None, 16, 112, 112, 3] is
+                used so you can evaluate any number of images at once
+        '''
+        self.config = config or C3DConfig.default()
+        self.sess = sess or tf.Session()
+        self.imgs = imgs or tf.placeholder(tf.float32, [None, 16, 112, 112, 3])
+        with tf.variable_scope('var_name') as var_scope:
```
brimoor: The point of `tf.variable_scope` is to give your variables informative names. For example:
```py
with tf.variable_scope("foo"):
    with tf.variable_scope("bar"):
        v = tf.get_variable("v", [1])
        assert v.name == "foo/bar/v:0"
```

brimoor: I don't see any reason to define the variables here. You should refactor this into the `build_c3d` method, and in there you should use scopes so that your variable names are well-organized.
(this isn't essential from an execution standpoint, but it's good manners.)

See `vgg16.py` for an example.

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+        stride = int(math.ceil(1.0 * num_frames / k))
+        rng = range(num_frames)
+        rng = list(rng[::stride])
+        if len(rng) < k:
+            rng.append(num_frames - 1)
+        elif len(rng) > k:
+            rng.pop()
+        assert len(rng) == k
+        with etav.VideoProcessor(inpath, out_size=out_size) as p:
+            for counter, img in enumerate(p):
```
brimoor: The proper way to only read a subset of frames is to pass the list of frames you want as the `frames` argument of `FFmpegVideoReader`.

Read the code you're calling before reimplementing something that already there!

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+class C3DConfig(Config):
+    '''Configuration settings for the C3D network.'''
+
+    def __init__(self, d):
+         self.model = self.parse_string(d, "model", default="C3D")
+         self.dropout = self.parse_number(d, "dropout", default="0.6")
+         self.batchsize = self.parse_number(d, "batchsize", default="1")
+         self.model_path = self.parse_string(d, "model_path", default="../eta/models/sports1m_finetuning_ucf101.model")
+         self.inpath = self.parse_string(d, "inpath", default="")
+         self.out_size = self.parse_number(d, "out_size", default="112")
```
brimoor: this can't be a parameter- your C3D implementation is hard-coded to 112 x 112 images. Right!?

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+        '''Builds a new C3D network
+
+        Args:
+            config: an optional C3DConfig instance. If omitted, the default
+                ETA configuration will be used
+            sess: an optional tf.Session to use. If none is provided, a new
+                tf.Session instance is created, and you are responsible for
+                scalling the close() method of this class when you are done
+                computing
+            imgs: an optional tf.placeholder of size [XXXX, 16, 112, 112, 3] to
```
brimoor: these are not images, they are video clips

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Convolution Layer
+        conv5 = self.conv3d('conv5a', pool4, _weights['wc5a'], _biases['bc5a'])
+        conv5 = tf.nn.relu(conv5, 'relu5a')
+        conv5 = self.conv3d('conv5b', conv5, _weights['wc5b'], _biases['bc5b'])
+        conv5 = tf.nn.relu(conv5, 'relu5b')
+        pool5 = self.max_pool('pool5', conv5, k=2)
+
+        # Fully connected layer
+        pool5 = tf.transpose(pool5, perm=[0,1,4,2,3])
```
brimoor: `perm=[0, 1, 4, 2, 3]`

eta/core/c3d.py
```diff
@@ -0,0 +1,340 @@
...
+
+    def __init__(self, d):
+         self.model = self.parse_string(d, "model", default="C3D")
+         self.dropout = self.parse_number(d, "dropout", default="0.6")
+         self.batchsize = self.parse_number(d, "batchsize", default="1")
+         self.model_path = self.parse_string(d, "model_path", default="../eta/models/sports1m_finetuning_ucf101.model")
+         self.inpath = self.parse_string(d, "inpath", default="")
+         self.out_size = self.parse_number(d, "out_size", default="112")
+         self.sample_method = self.parse_string(d, "sample_method", default="sliding_window")
+         self.frames = self.parse_number(d, "frames", default="16")
```
brimoor: this can't be a parameter- your C3D implementation is hard-coded to 16 frame clips. Right!?

eta/core/c3d.py
jasoncorso: And to how many frames?  Document this somewhere.

jasoncorso:
@jinyixin621 I will do a full pass once you have addressed Brian's comments.

eta/core/c3d.py
jinyixin621: Got it and set the ruler in Sublime Text.

eta/core/c3d.py
jinyixin621: Addressed.

eta/core/c3d.py
jinyixin621: Addressed, changed to embedding_frame_size

jinyixin621: Will make it a global constant since it is hard coded in C3D model

eta/core/c3d.py
jinyixin621: OK. I was using VideoProcessor before, don't see it can input a list. Yeah, FFmpegVideoReader makes life easier.

eta/core/c3d.py
jinyixin621: Addressed.

eta/core/c3d.py
jinyixin621: Addressed

eta/core/c3d.py
jinyixin621: Addressed

eta/core/c3d.py
jinyixin621: Addressed and still keep them as parameters to have more flexibilities

eta/core/c3d.py
jinyixin621: Addressed

eta/core/c3d.py
jinyixin621: Addressed.

eta/core/c3d.py
jinyixin621: Addressed and still keep them as parameters to have more flexibilities

eta/core/c3d.py
jinyixin621: As in the C3D tensorflow model the input is of size [batchsize, number_frames, img_size,  img_size, num_channels]. Should  I convert to the the dimension setting you mentioned and then reshape afterwards?

brimoor: I suppose you can leave it is `k x m x n` for now, but please document this in the docstrings.

eta/core/c3d.py
jinyixin621: Removed model_path but tested with this config locally.

eta/core/c3d.py
```diff
@@ -0,0 +1,381 @@
...
+    a numpy array of size [k, embedding_frame_siz, embedding_frame_size]
+
+    Args:
+        inpath: path to the input video
+        k: number of frames to extract for the input video
+        embedding_frame_size: dimension to use when embedding the frames
+    '''
+    data = []
+    num_frames = etav.get_frame_count(inpath)
+    assert k < num_frames
```
brimoor: you can write this whole mess as:
```py
rng = [int(round(i)) for i in np.linspace(1, num_frames, k)]
```
and it will even work in a reasonable way if there aren't enough frames

eta/core/c3d.py
```diff
@@ -0,0 +1,381 @@
...
+    np_arr_data = np.array(sampled_clips).astype(np.float32)
+    return np_arr_data
+
+
+class C3DConfig(Config):
+    '''Configuration settings for the C3D network.'''
+
+    def __init__(self, d):
+        self.model = self.parse_string(d, "model", default="C3D-UCF101")
+        self.dropout = self.parse_number(d, "dropout", default=0.6)
```
brimoor: never used

eta/core/c3d.py
```diff
@@ -0,0 +1,381 @@
...
+        embedding_frame_size: size of the output frames
+    '''
+    data = []
+    num_frames = etav.get_frame_count(inpath)
+    assert k < num_frames
+    with etav.FFmpegVideoReader(inpath, "1-%d" % k) as vr:
+        for img in vr:
+            img = etai.resize(img, embedding_frame_size, embedding_frame_size)
+            data.append(img)
+    np_arr_data = np.array(data).astype(np.float32)
```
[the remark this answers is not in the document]
jinyixin621: Addressed

eta/core/c3d.py
```diff
@@ -0,0 +1,381 @@
...
+from eta.core.features import Featurizer
+from eta.core.tfutils import TensorFlowModelWeights
+import eta.core.video as etav
+
+
+logger = logging.getLogger(__name__)
+
+
+def get_first_k_frames(inpath, k, embedding_frame_size):
+    '''Sample first k frames in a video. Return a numpy array of size
```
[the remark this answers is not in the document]
jinyixin621: Learned an addressed

eta/core/c3d.py
jinyixin621: Nice!

eta/core/c3d.py
```diff
@@ -0,0 +1,384 @@
...
+            layer = self.out
+        return self.sess.run(layer, feed_dict={self.clips: [clips]})[0]
+
+    def close(self):
+        '''Closes the TensorFlow session and frees up the network.'''
+        self.sess.close()
+        self.sess = None
+
+    def build_c3d(self, _X, _dropout):
+        with tf.variable_scope('var_name') as var_scope:
```
brimoor: this is still not quite what I requested: I wanted you to use `tf.name_scope()` to give a named scope to each layer. This command `tf.variable_scope('var_name')` is non-sensical.

See `eta.core.vgg16` to see how I'd like to see the model constructed with named layers.

jasoncorso:
What's the status here?  Shall I review or is further work being done?

brimoor:
Yixin has been tasked with verifying that our implementation here is "correct" by testing it on UCF-101 and verifying empirically that the predictions seem to be correct ~80% of the time, as expected

I found annoying subtle details about permuting a layer depending on which model was used, so we need confirmation that everything is working correctly.

PR #97, opened 2018-06-28
Small improvements to models registry

brimoor:
- adding a `publish_public_model` method to automate the process of publishing a public ETA model (assuming it has first been uploaded to Google Drive, which must be done by a Voxel51 admin)
- adding a `description` field to model entries
- updating VGG-16 model registry to include a description and informative filename

### DIFF
--- a/docs/models_dev_guide.md
+++ b/docs/models_dev_guide.md
@@ -62,12 +62,12 @@ The following is an example contents of a `manifest.json` file:
     "models": [
         {
             "base_name": "VGG-16",
-            "base_filename": "vgg16_weights.npz",
+            "base_filename": "vgg16-imagenet.npz",
             "version": null,
-            "description": "VGG-16 network originally trained by Frossard for the 1000 classes from ImageNet",
+            "description": "VGG-16 model trained on ImageNet. Source: https://github.com/ethereon/caffe-tensorflow",
             "manager": {
                 "config": {
-                    "google_drive_id": "0B7phNvpRqNdpT0lZU1NOWXIzRlE"
+                    "google_drive_id": "0B_-5gc091ngGMzBFMVJ0RE9HNmc"
                 },
                 "type": "eta.core.models.ETAModelManager"
             },
@@ -194,11 +194,15 @@ To publish a new model, you can follow the general receipe:
 
 ```py
 #
-# Full workflow for publishing a new model
+# Full workflow for publishing a new public ETA model
+#
+# Assumes the model has been uploaded to Google Drive by an ETA administrator
+# and you have the ID of the model in Google Drive
 #
 
 name = "your-model@1.0"
 description = "a short description of your model"
+google_drive_id = "XXXXXXXX"
 
 #
 # Recommend paths for the given model by looking for older versions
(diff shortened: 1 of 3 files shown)


docs/models_dev_guide.md
```diff
@@ -62,11 +62,12 @@ The following is an example contents of a `manifest.json` file:
     "models": [
         {
             "base_name": "VGG-16",
-            "base_filename": "vgg16_weights.npz",
+            "base_filename": "vgg16-imagenet.npz",
```
jasoncorso: Will the `flush` code handle the removal of the earlier model that now goes by a different name?  I suppose this may need to be done manually?

docs/models_dev_guide.md
```diff
@@ -189,38 +190,50 @@ automatically load a model given only its `name`. See the
...
-manager = ...
+#
+# The base filename (no version information, which is added automatically) and
+# models directory, respectively, to use when downloading the model. If you
+# are uploading a new version of an existing model, these arguments can be set
+# to None and the same values are inherited from the most recent version of the
+# model
+#
+base_filename = "your-model-weights.npz"
+models_dir = "/path/to/models/dir"
```
jasoncorso: As a reader of this doc, I see `models_dir`, I read the paragraph above `to use when downloading the model` and I think these thoughts:
1.  Isn't that models directory configured by all of eta?  Why do I need to specify it?  
2.  How will I know the root path to where people want to store their models?
3.  Is this also where it is stored in the eta public models repository?
4.  How do I tell the system where the current weights are?  Is this the same as `models_dir`?
Please clarify (if only to avoid you needing to answer these for everyone who uses it).

eta/core/models.py
```diff
@@ -255,16 +255,86 @@ def init_models_dir(new_models_dir):
     manifest.write_to_dir(new_models_dir)
 
 
-def recommend_paths_for_model(name):
-    '''Recommends a base filename and models directory for the given model
-    name, if possible.
+def publish_public_model(
```
jasoncorso: Same comment from the other PR about this style.  (Of course, take it or leave it as it meets the guidelines...)

eta/core/models.py
```diff
@@ -255,16 +255,86 @@ def init_models_dir(new_models_dir):
...
+    not be findable until you add the directory to your models search path.
+
+    Args:
+        name: a name for the model, which can optionally have "@<ver>" appended
+            to assign a version to the model
+        google_drive_id: the ID of the model file in Google Drive
+        description: an optional description for the model
+        base_filename: an optional base filename to use when writing the model
+            to disk. By default, a value is inferred as explained above
+        models_dir: an optional directory in which to register the model. By
```
jasoncorso: This should include the note that the `models_dir` is required for a first time, new model upload.

docs/models_dev_guide.md
brimoor: It must be done manually. We shouldn't make a habit out of doing this, I just wanted to get us off the ground with a convention to include the dataset in the filename for future clarity.

docs/models_dev_guide.md
brimoor: 1. I updated the code to fallback to the first directory on the model search path if necessary. This was definitely missing before.
4. This process doesn't involve the actual model file because it must have been previously uploaded to Google Drive (we shouldn't allow third-parties the ability to do that themselves)

For the rest, I added some more content to the dev guide. Hopefully it makes sense.

PR #104, opened 2018-07-13
Improving model class names

brimoor:
- renaming `eta.core.models.ModelWeights` to `PublishedModel` to better capture what the class is for
- also renaming `CanFeaturize.hint_featurize` to `force_featurize`, which is what it really does

### DIFF
--- a/eta/core/features.py
+++ b/eta/core/features.py
@@ -198,7 +198,16 @@ class CanFeaturize(object):
             decorator featurize_if_needed to always featurize.
         '''
         self.featurizer = featurizer
-        self.hint_featurize = hint_featurize
+        self.force_featurize = force_featurize
+
+    def __enter__(self):
+        if self.featurizer:
+            self.featurizer.start()
+        return self
+
+    def __exit__(self, *args):
+        if self.featurizer:
+            self.featurizer.stop()
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


eta/core/features.py
```diff
@@ -189,16 +189,26 @@ class CanFeaturize(object):
...
             featurizer: the actual featurizer to use when needed
-            hint_featurize: when True, this will force any input to the
-            decorator featurize_if_needed to always featurize.
+            force_featurize: whether to force any input to the
+                `featurize_if_needed` decorator to be featurized. By default,
+                this is False
         '''
         self.featurizer = featurizer
-        self.hint_featurize = hint_featurize
+        self.force_featurize = force_featurize
```
jasoncorso: I'm fine with this.  It is indeed more clear.

eta/core/features.py
```diff
@@ -189,16 +189,26 @@ class CanFeaturize(object):
...
+
+    def __enter__(self):
```
jasoncorso: Although this looks attractive, I do worry about whether this will get confusing for downstream developers as classes extending `CanFeaturize` will likely expose similar environment functionality for themselves.

brimoor: Upon second thought, I think this is the wrong place to implement this. I'm moving it to `Classifier`

PR #105, opened 2018-07-13
Vehicle pipeline updates

jeffreydominic:
* Updated the `sample_videos` and `resize_videos` modules to make them more robust
* Added some new types to `eta/core/types.py`


eta/core/types.py
```diff
@@ -489,6 +489,41 @@ def is_valid_path(path):
         )
 
 
+class NPZFile(File, ConcreteData):
```
brimoor: I don't think these are needed anymore as you updated the tracking modules to use directories rather than sequences of files like `/path/to/features/%05d.npz`. Can you please confirm?

eta/modules/sample_videos.py
```diff
@@ -56,8 +56,8 @@ class DataConfig(Config):
         input_path (eta.core.types.Video): The input video
 
     Outputs:
-        output_path (eta.core.types.ImageSequence): The output sampled video
-            frames
+        output_path (eta.core.types.VideoFile):
+            the path to the sampled video
```
brimoor: Changing the type of an established module's output is tricky, because now you've broken any existing pipelines or configs that counted on this generating an image sequence.

We were able to get away with this with the tracking modules because they were brand new and you knew all the configs to change, but here we need to make sure we retain backwards compatibility.

My recommendation was to add *another* output that optionally allows the video file to be written, not image frames. This is the easiest solution because its backwards compatible. But `output_path` is a bad name in this case; `output_impath` would be better...

eta/core/types.py
jeffreydominic: The Hungarian Matrix is stored as an `NPZFileSequence`.

brimoor: Oh I see. That's fine then.

eta/core/types.py
```diff
@@ -489,6 +489,42 @@ def is_valid_path(path):
         )
 
 
+class NpzFile(File, ConcreteData):
```
brimoor: @jeffreydominic note that I made this `NpzFile` instead of `NPZFile`, to match our existing conventions like `NpzModelWeights`. So you'll need to update the associated usages in THETA.

Sorry for the inconvenience.

jeffreydominic: That's fine. Should be any easy fix. @kunyilu It'll be good if you could add this fix to your `types.py` to avoid merge conflicts.

eta/modules/sample_videos.py
jeffreydominic: Right, I didn't think about that. However, changing `output_path` to `output_impath` would require changing the config files for other projects that use `sample_videos.py`. I'm fine with doing it, but I just want to check with you first. Do you think we should do this?

brimoor: Sorry I missed this comment earlier- yeah let's go ahead and use the best names we can. *I'll take care of it though*

brimoor:
@jeffreydominic I took over this PR a bit, so can you briefly review to make sure it looks good?

jeffreydominic:
Everything looks good. I tested the pipeline, after updating, and it works with these changes. I also added another type to `eta.core.types` (forgot to push that before). I'll also submit a quick PR for the updated pipeline.

PR #107, opened 2018-07-18
Add `index_in_frame` to DetectedObject

kunyilu:
- Add a new attribute `index_in_frame` to DetectedObject to inform the original index of an object in a frame

### DIFF
--- a/eta/core/objects.py
+++ b/eta/core/objects.py
@@ -91,6 +91,8 @@ class DetectedObject(Serializable, HasBoundingBox):
         score: (optional) an optional score for the object
         frame_number: (optional) the frame number in which this object was
             detected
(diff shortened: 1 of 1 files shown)


eta/core/objects.py
```diff
@@ -108,6 +108,7 @@ def __init__(
             score: (optional) an optional score for the object
             frame_number: (optional) the frame number in which this object was
                 detected
+            index_in_frame: (optional) the object index in one frame
```
brimoor: clarify this a bit: `the index of this object in the frame where it was detected`

kunyilu: done

kunyilu:
Addressed all comments.

PR #110, opened 2018-07-20
Call Serializable attributes

benjaminpkane:
In Serializable private attributes are excluded.

Feel free to clean this up, but this fix allows MATE to work.

Here is Serializable's attributes defintion

```python
    def attributes(self):
        '''Returns a list of class attributes to be serialized.
        Subclasses can override this method, but, by default, all attributes
        in vars(self) are returned, minus private attributes (those starting
        with "_").
        In particular, subclasses may choose to override this method if they
        want their JSON files to be organized in a particular way.
        '''
        return [a for a in vars(self) if not a.startswith("_")]
```

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -293,7 +293,8 @@ class BaseDataRecord(etas.Serializable):
         Returns:
             the list of attributes to be serialized
         '''
-        return [a for a in vars(self) if a not in self.excluded()]
+        attr = super(BaseDataRecord, self).attributes()
+        return [a for a in attr if a not in self.excluded()]
 
     def clean_optional(self):
         '''Deletes any optional attributes from the data record that are not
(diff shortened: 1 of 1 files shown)


eta/core/data.py
```diff
@@ -290,10 +290,14 @@ def attributes(self):
...
         serialized, i.e., all attributes that are not in `excluded()`
 
+        Note that Serializable.attributes() is called to remove private
+        attributes (those starting with "_").
+
         Returns:
             the list of attributes to be serialized
         '''
-        return [a for a in vars(self) if a not in self.excluded()]
+        attr = super(BaseDataRecord, self).attributes()
```
brimoor: why is this necessary now? Did BaseDataRecord previously remove private attributes?

PR #113, opened 2018-07-21
fixing a string issue for py3 support in the get_stream_info function.

jasoncorso:
This fix is necessary to use the `get_stream_info` function on Python3.  Confirmed to still work in Python2.


jasoncorso:
I added a few more fixes into this mix.  
Maybe I am the first to try these with Python 3?  This seems to be the source of these bugs.

PR #117, opened 2018-07-25
Making resize_videos and sample_videos modules more robust

brimoor:
This PR makes the `resize_videos` and `sample_videos` modules more robust and suitable for generic use as pre-processing steps 

Features include:
- ability to specify max frame size/frame rate to upper bound computation
- symlinking if no computation is required
- support for zipped inputs/outputs
- support for any output video type (video file, image sequence, image sequence directory)

A smattering of other small improvement are included, too.


eta/core/types.py
```diff
@@ -402,7 +402,7 @@ def gen_path(basedir, params):
 
     @staticmethod
     def is_valid_path(path):
-        return File.is_valid_path(path) and etav.is_supported_video(path)
+        return File.is_valid_path(path) and etav.is_supported_video_file(path)
```
jasoncorso: See comment above. Same.

eta/core/video.py
```diff
@@ -1229,20 +1246,28 @@ class FFmpegVideoSampler(FFmpeg):
...
-        sampler = FFmpegVideoSampler(24)
-        sampler.run("/path/to/video.mp4", "/path/to/frames/%05d.png")
+        sampler = FFmpegVideoSampler(fps=10)
+        sampler.run("/path/to/video.mp4", "/path/to/sampled.mp4")
     '''
 
-    def __init__(self, fps=None, **kwargs):
+    DEFAULT_OUT_OPTS = [
+        "-c:v", "libx264", "-pix_fmt", "yuv420p",
+        "-preset", "medium", "-crf", "23", "-an",
```
jasoncorso: 23 is a bit high, I think 20 or 21 will result in videos with less artifacts; if we need a default.

eta/modules/clip_videos.json
```diff
@@ -3,7 +3,7 @@
         "name": "clip_videos",
         "type": "eta.core.types.Module",
         "version": "0.1.0",
-        "description": "Module that generates clips from a video",
+        "description": "A module for generating clips from a video",
```
jasoncorso: Hmm; our convention is evolving.  Much of the codebase is "Class for" or "Pipeline for" or "Module ..."

eta/modules/sample_videos.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 '''
-Module for (re)sampling videos.
+A module for (re)sampling videos.
```
jasoncorso: This should be more thorough and clarify whether or not spatial resampling can be done (yes, via out_opts, I believe) in addition to temporal.

eta/core/features.py
```diff
@@ -330,9 +331,10 @@ def decorated(*args, **kwargs):
                             # it might be a file sequence.
                             #
                             # Currently the only such sequence we support is
-                            # a video, so we check if data is a valid video.
+                            # a video, so we check if data is a valid video
+                            # path.
                             #
-                            should_featurize = etav.is_valid_video(data)
+                            should_featurize = etat.Video.is_valid_path(data)
```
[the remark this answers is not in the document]
brimoor: that can't be accomodated- `is_valid_path` is an abstract method from `eta.core.types.Data` that all types must support for pipeline building to work.

What I did here was to make `should_featurize` be computed only from the path and not by actually firing up `ffmpeg`. That's the way the rest of the ETA works (e.g. the type checking that happens when a pipeline is built), so it's consistent to do the same here.

eta/core/types.py
brimoor: In our type system, `eta.core.types.Video` is an abstract type that code be an encoded video file `eta.core.types.VideoFile` or a sequence of images `eta.core.types.ImageSequence`.

`eta.core.video.SUPPORTED_VIDEO_FORMATS` only supports "video files" as defined above, so I updated the method names here to reflect that the related validation functionality is not handling image sequences.

In fact, I am also now renaming `SUPPORTED_VIDEO_FORMATS` to `SUPPORTED_VIDEO_FILE_FORMATS`, which should have also been done to drive home this distinction.

eta/modules/sample_videos.py
brimoor: `resize_videos` is for spatial and `sample_videos` is for temporal, although anything could be done in principle with `ffmpeg_out_opts`, yes.

Alas, these modules should probably be merged into one spatio-temporal module for efficiency... but I don't feel like doing it now.

brimoor: The first line of ETA module docstrings (the "short description" in Google docstring lingo) is shoved into the description field of the metadata file, so I have been keeping them short. Although we should brainstorm what exactly we want in these descriptions and what we'll do with them.

For the benefit of developers, I should also include a long description that discusses the module in more detail.

jasoncorso:
Ok, I understand the issue/challenge.

PR #161, opened 2018-08-28
Adding a field for different number of output

jinyixin621:
simple modification to support tunable class number of c3d model


brimoor:
Stale.

PR #180, opened 2018-12-30
Add UUID to schema

jasoncorso:
Adds lightweight support for the association of a unique id to all `AttributeSchema` classes, as discussed.  The uuid's are generated randomly and not tied to the naming of the Attribute by default, but this can be overridden.


brimoor:
Closing for now since AttributeSchema's are keyed by attribute name and thus they are already, by construction, unique identifiers.

PR #189, opened 2019-03-04
Adding interfaces for detectors and classifiers

brimoor:
In order to support deploying third-party object detectors into ETA pipelines, we need to officially declare our object detector interface. I also took the liberty of adding the analogous interface for classifiers.

### DIFF
--- a/eta/core/learning.py
+++ b/eta/core/learning.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python
+'''
+Core learning infrastructure.
+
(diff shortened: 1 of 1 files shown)


yashbhalgat:
Do you suggest I somehow run and check if this works correctly? If so, how?

Because, just by looking at it, it looks fine to me.

brimoor:
@yashbhalgat thanks for reviewing carefully. There's really no way to test this. Just a definition of object detector/classifier interfaces. Just approve :)

PR #191, opened 2019-03-05
Added cull_with_function method to DataRecords

mattphotonman:
This will be useful in theta development.  Also tweaked the cull method.

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -827,9 +827,31 @@ class DataRecords(DataContainer):
                 "Either keep_values or remove_values must be provided")
 
         # Cull records
-        inds = []
+        inds = set()
         for v in keep_values:
-            inds += lud[v]
+            inds.update(lud[v])
+        self.keep_inds(inds)
+
+        return len(self)
+
+    def cull_with_function(self, field, func=lambda x: x):
+        '''Cull records from the container for which `field` returns
+        something that evaluates to False when passed through func.
+
+        Args:
+            field: the field to process
+            func: the test function
+
+        Returns:
+            the number of elements in the container
+        '''
+        lud = self.build_lookup(field)
+
+        # Cull records
+        inds = set()
+        for v in lud:
+            if func(v):
+                inds.update(lud[v])
         self.keep_inds(inds)
 
         return len(self)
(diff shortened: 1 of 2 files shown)


eta/core/data.py
```diff
@@ -827,9 +827,31 @@ def cull(self, field, keep_values=None, remove_values=None):
                 "Either keep_values or remove_values must be provided")
 
         # Cull records
-        inds = []
+        inds = set()
```
jasoncorso: It's not clear to me why you changes inds from a list to a set.  The `self.keep_inds` function from `Container` expects a list.  (Actually, that function in line 440 of `core/serial.py` now converts the input to a set.  Perhaps the interface to keep_inds be redefined to require a set in its contract and this conversion should only happen once.  Is a `set` more efficient for the list comprehension in `keep_inds`?)

brimoor: `set` is more efficient both here and in `keep_inds` because sets are implemented as hash maps with `O(1)` lookups.

Strictly speaking, there are many places in our code where we document that the arg should be a `list` when we all we do is iterate over it and thus it could be any `iterable`. Not requesting any action on this front, just commenting.

eta/core/data.py
```diff
@@ -827,9 +827,31 @@ def cull(self, field, keep_values=None, remove_values=None):
...
         for v in keep_values:
-            inds += lud[v]
+            inds.update(lud[v])
+        self.keep_inds(inds)
+
+        return len(self)
+
+    def cull_with_function(self, field, func=lambda x: x):
```
brimoor: why provide a default value here? doesn't make sense to call this method without providing a function.

We are not one of those shops that requires almost every arg to have default values so that we can "silently" deprecate function syntaxes without breaking existing code. I generally like implementations that minimize the number of keystrokes I make ;)

eta/core/data.py
mattphotonman: Agreed.  I propose changing the doc string for `inds` to specify an `iterable` rather than `list`, and will make that change unless there are any objections.

eta/core/data.py
mattphotonman: Fixed.

PR #192, opened 2019-03-06
Added to_list method to Container

mattphotonman:
One more change to eta.


mattphotonman:
I retract this pull request since this can be accomplished using the `__iter__` method of `Container`.

PR #193, opened 2019-03-06
Data records fix

brimoor:
```py
import sys

import eta.core.data as etad

sys.path.insert(1, "/path/to/theta")
import core.vehicles as thev

dr = etad.DataRecords(thev.VehicleDataRecord)
vdr = thev.VehicleDataRecord("Honda", "Accord", "red", "2004", "sedan", "")
dr.add(vdr)

# this should work
str1 = dr.to_str()  # not reflective
dr1 = etad.DataRecords.from_str(str1, record_cls=thev.VehicleDataRecord)

# this previously didn't work, but now does work
str2 = dr.to_str(reflective=True)  # reflective
dr2 = etad.DataRecords.from_str(str2)
```

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -853,19 +853,23 @@ class DataRecords(DataContainer):
         Args:
             d: a DataRecords dictionary
             record_cls: an optional records class to use when parsing the
-                records dictionary. If not provided, the DataRecord dictionary
-                must define it
+                records dictionary. If not provided, the DataRecords dictionary
+                you are loading must have been written in reflective mode
 
         Returns:
             a DataRecords instance
         '''
-        rc = record_cls or d.get(cls._ELE_CLS_FIELD, None)
-        if rc is None:
-            raise DataRecordsError(
-                "Need record_cls to parse the DataRecords dictionary")
+        if record_cls is None:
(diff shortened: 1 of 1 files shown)


eta/core/data.py
```diff
@@ -853,35 +853,24 @@ def from_dict(cls, d, record_cls=None):
...
+                    "populated, so you must manually specify the `record_cls` "
+                    "to use when loading it" % cls._ELE_CLS_FIELD)
+            record_cls = etau.get_class(record_cls_str)
 
         return DataRecords(
-            record_cls=rc,
-            records=[rc.from_dict(r) for r in d[cls._ELE_ATTR]])
-
-    @classmethod
-    def from_json(cls, json_path, record_cls=None):
```
mattphotonman: You removed the from_json method because it is automatically defined via from_dict, right?

brimoor: I removed it because the inherited method `Serializable.from_json()` passes it onto `from_dict` already. Everything still works without this implementation here.

jasoncorso: Noting that (1) I believe this was created before the `from_json` handled named args robustly (not important, but just noting) and (2) how do we document this case -->  the `from_json` that is now used does not explicitly have the `record_cls` argument and hence how is the new developer going to know this can be called in this way.

PR #194, opened 2019-03-06
Only including top-level reflective attributes in Containers

brimoor:
Previously the "reflective" option for Containers embedded the reflective attributes recursively for all Container elements. However, this is unnecessary as Containers store homogeneous lists of elements. Therefore, I updated the `reflective=True` option to only include the `_ELE_CLS` once at the top-level. This saves space while maintaining generality.

Example:

```py
import eta.core.data as etad

dr = etad.DataRecords(record_cls)

print(dr.to_str())

'''
{
    "records": []
}
'''

print(dr.to_str(reflective=True))
'''
{
    "_CLS": "eta.core.data.DataRecords",
    "_RECORD_CLS": "<record_cls>",
    "records": []
}
'''
```

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -846,6 +846,29 @@ class DataRecords(DataContainer):
         '''
         return self.extract_inds(indices)
 
+    def attributes(self):
+        '''Returns a list of class attributes to be serialized.'''
+        return [self._ELE_ATTR, self._ELE_CLS_FIELD]
+
+    def serialize(self, reflective=False):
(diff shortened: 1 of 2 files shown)


brimoor:
@mattphotonman thanks for the comments. ready for final review now

brimoor:
I addressed that by adding `**kwargs: optional keyword arguments for self.from_dict()` to the docstrings of the Serializable classes. Now either the reimplemented methods have docstrings or they inherit methods from Serializable with the appropriate documentation.

PR #201, opened 2019-03-12
Added HtmlFile and TextFile to core.types

mattphotonman:


eta/core/types.py
```diff
@@ -1077,3 +1077,35 @@ class PickleFileDirectory(Directory):
...
+    @staticmethod
+    def gen_path(basedir, params):
+        return os.path.join(basedir, "{name}.txt").format(**params)
+
+    @staticmethod
+    def is_valid_path(path):
+        return File.is_valid_path(path) and etau.has_extension(path, ".txt")
+
+
+class HTMLFile(File, ConcreteData):
```
brimoor: FYI- I capitalized `HTMLFile`. Our convention is to capitalize all letters in abbreviations. For example we have `JSONFile`.

I was about to comment this on your previous PR for `NpzFile`, but then I realized that we seem to have broken this convention for all `Npz*` things. Oops! But let's stick to `HTMLFile` here.

PR #202, opened 2019-03-13
Added NpzWriteable and FeatureArray classes

mattphotonman:


eta/core/data.py
```diff
@@ -1029,3 +1031,11 @@ def optional(cls):
     @classmethod
     def required(cls):
         return ["video_path", "label"]
+
+
+class FeatureArray(NpzWriteable):
```
jasoncorso: The semantics of FeatureArray are not exact enough for me.  `y` is not features.   `X` could plausibly be raw data.  Does this name need to change?

jasoncorso: Second comment for discussion.
Do we want to be so tied to Npz?  While sufficient for now, I think a broader approach may be preferred here; perhaps where this inherits a contract and the writer/format is "connected" in some way.

mattphotonman: @jasoncorso is this just a question of the name of the class, or somehow clarifying what `y` and `X` actually are?  For the record, @brimoor 's description is accurate:
```
X is an n x d array whose rows contain features
y is a length-n array of labels
```

mattphotonman: Are there any classes currently used to store images as numpy arrays, for example, that I am missing and could be generalized to include this use case?

eta/core/data.py
```diff
@@ -1029,3 +1031,11 @@ def optional(cls):
...
+    '''Class representing a feature array `X` and corresponding labels `y`.'''
```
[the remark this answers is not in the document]
mattphotonman: done

eta/core/data.py
brimoor: There are not any existing classes that encapsulate binary-type data in "official" formats. We store video and images directly as `.mp4`, `.png`, etc.

Jason's is suggesting that we may consider abstracting this so modules could work with labeled features stored in non-npz formats with the same interface. This is interesting but I would personally wait for the need before acting on it.

brimoor: The goal is just to make the name + documentation more descriptive of what we actually intend to store in this class. I'm fine with keeping the scope of this type to a collection of labeled feature vectors (which could be vectorized raw images, if desired)

PR #204, opened 2019-03-20
Ability to extract keyframes only from a video

mattphotonman:
This is a very fast way to get a sample of frames from a video if you don't care about the exact index of the frames.

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1950,7 +1950,8 @@ class FFmpegVideoReader(VideoReader):
         self._stream_info = VideoStreamInfo.build_for(inpath)
         self._ffmpeg = FFmpeg(
             in_opts=[
-                "-skip_frame", "nokey"
+                "-skip_frame", "nokey",
+                "-vsync", "0"
             ] if keyframes_only else None,
             out_opts=[
                 "-f", 'image2pipe',         # pipe frames to stdout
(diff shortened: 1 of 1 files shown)


eta/core/video.py
```diff
@@ -1625,6 +1654,7 @@ def __init__(
             self,
             inpath,
             frames=None,
+            keyframes_only=False,
```
brimoor: I renamed this to `keyframes_only` for consistency with the `frames` argument (its not `in_frames`).

Note also that I was bold and moved this kwarg up in the list. You had it at the end. We are diligent about using `key=value` syntax whenever using keyword arguments, so this is generally safe to do.

eta/core/video.py
```diff
@@ -1639,6 +1669,9 @@ def __init__(
             inpath: path to the input video. Passed directly to a VideoReader
             frames: an optional range(s) of frames to process. This argument
                 is passed directly to VideoReader
+            keyframes_only: whether to only extract keyframes when reading the
```
brimoor: I clarified that you can use `frames` and `keyframes_only` together, which is cool!

eta/core/video.py
```diff
@@ -1973,6 +2020,11 @@ def _grab(self):
             return False
 
     def _retrieve(self):
+        # Stop when ffmpeg returns empty bits, meaning it has gone past the end
+        # of the video
+        if not self._raw_frame:
```
brimoor: thanks for adding this too

mattphotonman:
> LGTM. Tested and seems to work. Am I correct in assuming that one can use `keyframes_only` and `frames` together to, for example, extract the first and third keyframes?
> 
> I tested and it appears so. Cool!

Yeah, that's right.

PR #207, opened 2019-03-24
Infrastructure for deploying classifiers

brimoor:
This PR does the following things:
- generalizes the interface for classifiers in `eta.core.learning`
- improves the documentation in `eta.core.features`
- fixes some broken windows that I found in the code along the way

Two new classifier interfaces are declared:
- `eta.core.learning.ImageClassifier`: a classifier that classifies single images (or processes an image tensor but returns predictions for each image)
- `eta.core.learning.VideoFramesClassifier`: a classifier that takes a tensor of images and returns a single prediction for the tensor

These abstractions were necessary in order to be concrete enough that modules can know what to do with the classifiers that they are provided (e.g., does the classifier predict on tensors directly or does it classify each image individually?.

In addition, a meta-classifier called `VideoFramesVotingClassifier` that uses an `ImageClassifier` to classify each frame of a video tensor and then majority votes on each attribute to determine the final prediction for the tensor.

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -1,6 +1,10 @@
 '''
 Core utilities for rendering annotations on media.
 
+@todo support for rendering video-level attributes
+@todo support for rendering frame-level attributes
+@todo support for rendering object-level attributes
+
 Copyright 2019, Voxel51, Inc.
 voxel51.com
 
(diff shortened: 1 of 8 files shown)


eta/core/annotations.py
```diff
@@ -1,6 +1,10 @@
 '''
 Core utilities for rendering annotations on media.
 
+@todo support for rendering video-level attributes
```
jasoncorso: Questioning whether we plan to keep functionality like this in ETA.

brimoor: The goal is for ETA users to be able to render a  `VideoLabels` instance on a video, which should help popularize ETA formats.

yashbhalgat:
@mattphotonman and I had a look, LGTU! Merging it.

yashbhalgat:
@mattphotonman and I had a look, LGTU! Merging it.

PR #210, opened 2019-04-01
Visualize frame attributes in a video (visualize_labels module)

yashbhalgat:
Editing the `visualize_labels` module in `eta/core` to support frame-level attribute visualizations.


yashbhalgat:
@brimoor , I have modified the module so that it also displays the attributes now.
Have a look at this screenshot of the output of an end-to-end Deep Classifier pipeline. :)
Selection_016

PR #211, opened 2019-04-02
Adding support for rendering frame level attributes in ETA

brimoor:
Example:

Screen Shot 2019-04-02 at 12 25 00 AM

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -162,7 +162,7 @@ _DEFAULT_LOGO = etal.Logo.load_default()
 def annotate_video(
         input_path, video_labels, output_path, add_logo=True, logo=None,
         colormap=None, font=None, alpha=None, text_color=None, pad=None,
-        linewidth=None):
+        linewidth=None, attrs_bg_color=None, attrs_gap=None):
     '''Annotates the video with the given labels.
 
     Args:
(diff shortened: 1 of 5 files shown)


eta/core/annotations.py
```diff
@@ -1,11 +1,7 @@
 '''
 Core utilities for rendering annotations on media.
 
-@todo support for rendering video-level attributes
```
brimoor: @yashbhalgat note that I went ahead and supported the other two todo items

eta/core/models.py
```diff
@@ -927,13 +926,7 @@ class PickledModel(PublishedModel):
     '''Class that can load a published model stored as a .pkl file.'''
 
     def _load(self):
-        try:
```
brimoor: @yashbhalgat note that I moved this `try-except` to a `eta.core.serial.read_pickle`. The idea is that we only want to implement reading pickle files once and then re-use it everywhere. You fixed this case, but there was a duplicate implementation of pickle reading in `Picklable.from_pickle`, so I fixed that.

eta/core/annotations.py
yashbhalgat: Yes.. This looks good now

eta/core/models.py
yashbhalgat: Good idea!

jasoncorso:
Note that my preference for the video level attributes is for the box to
extend to the top and left boundaries of the frame.  Seems cleaner to me.

brimoor:
I matched how it looks in player51. it is customizable in this code however- one can change paddings as desired

PR #215, opened 2019-04-05
added function for central crop

yashbhalgat:
@brimoor , I guess I forgot to make a PR for this yesterday.

### DIFF
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -702,11 +702,18 @@ def resize(img, width=None, height=None, *args, **kwargs):
 
 
 def central_crop(img, h, w):
-    '''Crops the central part of an image and outputs an image of shape (h, w)
-    Works for both 2D and 3D images.
+    '''Crops the central part of an image of required size. Works for both
+    2D and 3D images.
+
+    Args:
+        img: input image
+        (h, w): required output height and width
+
+    Returns:
+        A cropped portion of the image of height `h` and width `w`.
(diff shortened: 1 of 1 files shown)


eta/core/image.py
```diff
@@ -701,6 +701,17 @@ def resize(img, width=None, height=None, *args, **kwargs):
     return cv2.resize(img, (width, height), *args, **kwargs)
 
 
+def central_crop(img, h, w):
+    '''Crops the central part of an image and outputs an image of shape (h, w)
```
brimoor: For top-level public functions, we prefer to use a full docstring in Google python style:
```py
'''Description here.

Args:
    img: xxx
    (h, w): yyy

Returns:
    description here
'''
```

eta/core/image.py
```diff
@@ -701,6 +701,17 @@ def resize(img, width=None, height=None, *args, **kwargs):
...
+    Works for both 2D and 3D images.
+    '''
+    bounding = (h, w)
+    start = tuple(map(lambda a, da: a//2-da//2, img.shape, bounding))
```
brimoor: linting: one space before/after punctuation. `a // 2 - da // 2`

eta/core/image.py
yashbhalgat: Oh, didn't note this. I just followed the docstring format of the function above this one `resize`.

PR #219, opened 2019-04-19
Modifying storage.py to support recursive download of files

yashbhalgat:
* Modified the query in `list_files_in_folder` function to exclude trashed files.
* Modified the `download_files_in_folder` function to support recursive download of files.
* The `import operator` line in `image.py` is a bug fix.

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -688,7 +688,7 @@ class GoogleDriveStorageClient(StorageClient, NeedsGoogleCredentials):
         # Build file list
         files = []
         page_token = None
-        query = "'%s' in parents" % folder_id
+        query = "'%s' in parents and trashed=false" % folder_id
         while True:
             # Get the next page of files
             response = self._service.files().list(
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -765,7 +765,7 @@ class GoogleDriveStorageClient(StorageClient, NeedsGoogleCredentials):
 
     def download_files_in_folder(
             self, folder_id, local_dir, skip_failures=False,
-            skip_existing_files=False):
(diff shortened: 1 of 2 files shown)


yashbhalgat:
@brimoor @mattphotonman please take a look at this when you have time. :)

eta/core/storage.py
```diff
@@ -805,13 +805,21 @@ def download_files_in_folder(
...
             file_id = f["id"]
             try:
-                local_path = os.path.join(local_dir, filename)
-                with etau.Timer() as t:
-                    self.download(file_id, local_path)
-                    filenames.append(filename)
-                logger.info(
-                    "File '%s' downloaded to '%s' (%s) (%d/%d)", filename,
-                    local_path, t.elapsed_time_str, idx, num_files)
+                if recursive and self.is_folder(file_id):
```
brimoor: calling `is_folder` here will be quite slow because you are calling it for every file, and Google Drive API requests are slow. Instead, you should be able to modify `list_files_in_folder` to include whether the object is a file or folder. Then you wouldn't need all these extra requests here.

eta/core/storage.py
```diff
@@ -804,14 +806,24 @@ def download_files_in_folder(
...
             try:
-                local_path = os.path.join(local_dir, filename)
-                with etau.Timer() as t:
-                    self.download(file_id, local_path)
-                    filenames.append(filename)
-                logger.info(
-                    "File '%s' downloaded to '%s' (%s) (%d/%d)", filename,
-                    local_path, t.elapsed_time_str, idx, num_files)
+                if (recursive and
+                    file_type=="application/vnd.google-apps.folder"):
```
brimoor: style nit: one space before and after `==`

Issue #220, opened 2019-04-25
Use `vsync 0` when reading videos with ffmpeg


brimoor:
We should consider always using `vsync 0` when reading video via `ffmpeg`.

It has been observed that, for certain high frame rate videos, frames may be dropped without `vsync 0` as they may have the same timestamps.

As a library, ETA needs a global contract for converting `frame number <-> timestamp <-> img`.

brimoor:
https://github.com/voxel51/eta/pull/226
https://github.com/voxel51/eta/pull/228

PR #226, opened 2019-05-16
Adding -vsync 0 to default ffmpeg options

brimoor:
I haven't been able to find a video that doesn't work without `-vsync 0` , but I went ahead and added it as a default option anyway.

The ffmpeg on my laptop has no problem with a 60fps BDD video before or after this change.

Addresses https://github.com/voxel51/eta/issues/220.

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1423,13 +1423,13 @@ def extract_clip(
     # frames of the clip will be exactly the same as those encountered via
     # other clip-based methods in ETA?
     #
-    in_opts = ["-vsync", "0"]
+    in_opts = []
     if start_time is not None:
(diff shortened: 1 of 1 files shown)


eta/core/video.py
```diff
@@ -1993,6 +1981,7 @@ def __init__(self, inpath, frames=None, keyframes_only=False):
         self._ffmpeg = FFmpeg(
             in_opts=in_opts,
```
mattphotonman: I noticed above that had previously added the `-vsync 0` option as an _input option_ when the `keyframes_only` option is turned on.  I'm wondering if there are any other circumstances where this would be appropriate as an input option?

PR #227, opened 2019-05-16
added stride parameter to sample_first_frames

yashbhalgat:
Possibly less effective, but much faster than `sliding_frames_sampling`.

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1530,7 +1530,7 @@ def sample_select_frames(video_path, frames, output_patt=None, fast=False):
         return imgs
 
 
-def sample_first_frames(arg, k, size=None):
+def sample_first_frames(arg, k, stride=2, size=None):
     '''Samples the first k frames in a video.
 
     Args:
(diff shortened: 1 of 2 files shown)


eta/core/video.py
```diff
@@ -1530,7 +1530,7 @@ def sample_select_frames(video_path, frames, output_patt=None, fast=False):
         return imgs
 
 
-def sample_first_frames(arg, k, size=None):
+def sample_first_frames(arg, k, stride=2, size=None):
```
brimoor: you're changing the default behavior! should be `stride=1`

brimoor: need to document new parameter in docstring

eta/core/video.py
```diff
@@ -1546,11 +1546,12 @@ def sample_first_frames(arg, k, size=None):
     # Read frames ...
     if isinstance(arg, six.string_types):
         # ... from disk
-        with FFmpegVideoReader(arg, frames="1-%d" % k) as vr:
+        with FFmpegVideoReader(
```
brimoor: needs to be `range(1, stride * k  + 1, stride)`

eta/core/video.py
```diff
@@ -1546,11 +1546,12 @@ def sample_first_frames(arg, k, size=None):
...
+            arg, frames=[i for i in range(1, stride*k, stride)]) as vr:
             imgs = [img for img in vr]
     else:
         # ... from tensor
-        imgs = arg[:k]
+        imgs = arg[:k*stride:stride]
```
brimoor: style: one space before and after `*`. Optional `()` for clarity: `(k * stride)`

eta/core/video.py
yashbhalgat: Oh right, my bad.

PR #230, opened 2019-05-21
Google API retry on certain errors

tylerganter:
Following google's recommendations for handling API errors with retries. Two error types are attempted with exponential backoff, one with constant 30 second delay. All with max 10 attempts, which is the one somewhat arbitrary value selected.

Uses the python `retrying` package.


eta/core/storage.py
```diff
@@ -58,6 +60,35 @@
 logging.getLogger("paramiko").setLevel(logging.ERROR)
 
 
+def google_cloud_api_retry(func):
```
brimoor: I bet this could also be applied to `GoogleDriveStorageClient`

tylerganter: I just applied it to all functions in google drive, noting that it only catches specific errors and otherwise does nothing, so it's pretty safe to just apply it anywhere :)

brimoor: Did you actually test to see if the Google Drive API will raise errors from these namespaces?

brimoor: Also, importantly, some of these functions are not purely google API calls. They may perform multiple. For example, if you retry an entire function that tries to upload a list of files in series when the last one fails, then you'll get multiple copies of some files.

It's only safe to retry single API calls. Or actions that are otherwise idempotent.

tylerganter:
I may have gotten ahead of myself. So if something similar is applied to Google Drive it should only be applied to functions that don't call other exposed functions (such as `upload_files_in_folder`).

[This](https://developers.google.com/api-client-library/python/guide/media_upload) seems like a good start, but also not something I want to dive into right now

PR #240, opened 2019-06-11
ImageMetadata.build_for() fix for single channel images

MikeJeffers:
The shape tuple does not have a 3rd parameter when it is just a 2D matrix, as is the case for single channel images.
Returns 1 in this case to correctly report the number of channels.


MikeJeffers:
My bad this is fixed with recent changes on develop.

PR #249, opened 2019-07-09
Converting video datasets to image datasets

mattphotonman:
This PR refactors the `to_image_set()` method of `eta.core.datasets.LabeledVideoDataset` into a separate top-level function `sample_videos_to_images()`.  This new function now supports pre-filtering frames by their labels, and fixing either the number of images in the output image dataset or the stride.


brimoor:
I defer to Mike and Ben for review. BUT let me say: yay for export!

PR #252, opened 2019-07-10
Datasets: builder options as transforms

MikeJeffers:
Builder should build with a more flexible collection of compliant transformer objects

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1018,7 +1018,6 @@ class LabeledDatasetError(Exception):
     pass
 
 
-
 ################################################################################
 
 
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1071,98 +1069,112 @@ class LazyLabeledVideoEntry(LazyLabeledDataEntry):
...
 
-        #@TODO build me!
-        return LabeledDataset(None)
+    def transform(self, input):
+        # @TODO implement Balancing!!
+        return input
 
 
-class LabeledVideoDatasetBuilder(LabeledDatasetBuilder):
+class Filter(DatasetTransformer):
```
brimoor: small comment on naming: `Filter` is so generic. `DatasetFilter` would be preferable. Same for other classes

MikeJeffers: Yes - will fix that

PR #253, opened 2019-07-10
Dataset-transforms: Filter

MikeJeffers:
First pass, and some refactor

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1069,22 +1069,25 @@ class LazyLabeledVideoEntry(LazyLabeledDataEntry):
      in memory!
     '''
 
-    def __init__(self, data_path, label_path, duration, start_frame, end_frame,
-                 total_frame_count):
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1141,9 +1144,30 @@ def __init__(self, schema):
...
+                except etad.AttributeContainerSchemaError as err:
+                    logger.warn(err)
+        return segment
+
+    def transform(self, src):
+        for entry in src:
+            labels = entry.get_labels()
+            labels = self._extract_video_labels(entry.start_frame,
+                                                entry.end_frame, labels)
+            entry.set_labels(labels)
```
benjaminpkane: Very nice!

benjaminpkane: I do think that actual filtering should only happen once in `build`. Passing multiple  filters should just add to the schema that will be used by `build` to filter at the end.

PR #255, opened 2019-07-11
Dataset builder merge

benjaminpkane:
Merge `LazyLabeledDataset` in `LabeledDatasetBuilder`.

The transformers now operate on the builder, which feels somewhat clunky.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1193,7 +1193,7 @@ class LabeledDatasetBuilder(object):
 
     def __init__(self):
         self.transformers = []
-        self._dataset = None
+        self._dataset = BuilderDataset(schema=schema)
 
     def add_record(self, record):
         self.dataset.add(record)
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1219,14 +1178,36 @@ class LabeledDatasetBuilder(object):
...
+        return self.entries
 
     def add_transform(self, transform):
         self.transformers.append(transform)
 
     def build(self):
         for transformer in self.transformers:
-            self.dataset = transformer.transform(self.dataset)
-        return self.dataset.make_dataset()
+            transformer.transform(self)
```
MikeJeffers: Oh transformers take the builder in directly... hmm 
This is why putting too much responsibility in one object can be problematic, this could allow a transformer to add itself to the builder.  I think there should be a POJO (or POPO) that simply has the collection or records.  But make the interface for its construction and lifecycle internal to that of the builder:
 - Make builder
 - Make records
 - Add records to builder -> builder adds record to internal object
 - Builder passes internal collection to transformers, not self

benjaminpkane: Right. I've made some changes.

eta/core/datasets.py
```diff
@@ -1111,27 +1077,35 @@ def __init__(self, video_path, labels_path, start_frame=None,
...
+    DatasetTransformers operate on BuilderDatasets.
+    '''
+
+    def __init__(self, schema=None):
+        self.records = []
+        self.schema = schema
+
+    def add(self, record):
+        self.records.append(record)
+
```
MikeJeffers: needs `get_entries` etc.

eta/core/datasets.py
```diff
@@ -1190,12 +1163,11 @@ def _extract_video_labels(self, start_frame, end_frame, labels):
         return segment
 
     def transform(self, src):
-        for entry in src.get_entries():
-            labels = entry.get_labels()
-            labels = self._extract_video_labels(entry.start_frame,
-                                                entry.end_frame, labels)
-            entry.set_labels(labels)
-        return src
+        for record in src.get_entries():
```
MikeJeffers: BuilderDataset needs some method for allowing transformers to access it's data - and manipulate, replace, remove, and add to it

eta/core/datasets.py
benjaminpkane: Added!

PR #256, opened 2019-07-11
Build from a LabeledDataset

benjaminpkane:
To run a series of transforms on a `LabeledDataset` we call `builder()` on the dataset to create the builder instance. You then add transforms as you would expect.

No output yet, so this has not been heavily tested.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -277,21 +277,21 @@ class LabeledDatasetBuilder(object):
     '''
 
     def __init__(self, schema=None):
-        self.transformers = []
+        self._transformers = []
         self._dataset = None
 
     def add_record(self, record):
         self._dataset.add(record)
 
     def add_transform(self, transform):
-        self.transformers.append(transform)
+        self._transformers.append(transform)
 
     @property
     def record_cls(self):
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -788,6 +839,12 @@ def _ensure_empty_dataset_dir(dataset_path):
         etau.ensure_dir(data_subdir)
         etau.ensure_dir(labels_subdir)
 
+    def builder(self):
```
benjaminpkane: This is the function users would call. perhaps `transformer`? Naming is hard.

eta/core/datasets.py
```diff
@@ -1076,30 +1159,52 @@ def __init__(self, video_path, labels_path, start_frame=None,
...
+        attrs += [
+            "start_frame",
+            "end_frame",
+            "duration",
+            "total_frame_count"
+        ]
+        return attrs
+
+
+class BuilderDataset(etad.DataRecords):
```
benjaminpkane: Note that this is a `DataRecords` instance which are meant to hold `BaseDataRecord`s. `add` and much more included!

PR #258, opened 2019-07-11
Basic output from dataset builder

benjaminpkane:


eta/core/datasets.py
```diff
@@ -1136,6 +1151,17 @@ def __init__(self, image_path, labels_path):
         super(BuilderImageRecord, self).__init__(image_path, labels_path)
         self.labels_cls = etai.ImageLabels
 
+    def build(self, dir_path, filename, pretty_print=False):
+        data_filename = filename + os.path.splitext(self.data_path)[1]
+        data_path = os.path.join(dir_path, data_filename)
+        etau.copy_file(self.data_path, data_path)
```
MikeJeffers: Will the file be on a local disk path at this time?

eta/core/datasets.py
```diff
@@ -1159,11 +1186,37 @@ def __init__(self, video_path, labels_path, start_frame=1,
...
-        self.end_frame = video_metadata.total_frame_count
+    def _init_from_video_metadata(self):
+        self.total_frame_count = self.metadata.total_frame_count
+        self.duration = self.metadata.duration
+        self.end_frame = self.metadata.total_frame_count
+
+    @property
+    def metadata(self):
+        if not self._metadata:
+            self._metadata = etav.VideoMetadata.build_for(self.data_path)
```
MikeJeffers: We avoided loading the metadata file directly elsewhere, will this require the file, or can this be constructed optionally from a provided dictionary of metadata fields?

benjaminpkane: This can be avoided. I'll update this.

eta/core/datasets.py
benjaminpkane: Yes, via `validated_path`.

PR #259, opened 2019-07-11
updated docstring for write_manifest()

rpinnaka:


eta/core/datasets.py
```diff
@@ -390,6 +391,24 @@ def set_description(self, description):
     def write_manifest(self, filename, description=None):
         '''Writes the manifest to a new file inside the base dataset directory.
 
+        Manifest is stored on disk in the following format
```
brimoor: It would be preferable to move this example JSON to the docstring of the `LabeledDatasetIndex` class. Then everyone who uses manifests and wants to know what they look like can find the info there.

The docstrings of the `LabeledImageDataset` and `LabeledVideoDataset` classes already specify that their manifests are stored in `LabeledDatasetIndex` format, so the relevant pointers should already be in place

rpinnaka:
yes @brimoor, it would be more meaningful to update the docstring of the `LabeledDatasetIndex` class instead. Updated accordingly, you can please re-review it. Thanks.

rpinnaka:
Thanks @mattphotonman, for mentioning that too. Updated it.

mattphotonman:
Thanks @rpinnaka , LGTM!  Please merge when ready.

PR #260, opened 2019-07-12
LabeledDatasetBuilder

benjaminpkane:
Ready for review now that `Balancer` is completed!


eta/core/datasets.py
```diff
@@ -1338,3 +1373,839 @@ def required(cls):
...
+            A: the occurrence matrix
+            b: the target vector to match
+
+        Returns:
+            x: the solution vector. x[j]=1 -> omit the j'th record
+        '''
+        best_x = np.zeros(A.shape[1], dtype=np.dtype('int'))
+        best_score = self._solution_score(b - np.dot(A, best_x))
+
+        while True:
```
brimoor: I haven't translated to math to understand what you're doing, but `while True` scares me

eta/core/datasets.py
```diff
@@ -1338,3 +1373,839 @@ def required(cls):
...
+
+        Returns:
+            x: the solution vector. x[j]=1 -> omit the j'th record
+        '''
+        best_x = np.zeros(A.shape[1], dtype=np.dtype('int'))
+        best_score = self._solution_score(b - np.dot(A, best_x))
+
+        import random
+        random.seed(1)
+        for _ in range(10000):
```
brimoor: always performing 10K O(n^2) operations? yikes

I haven't translated what you're doing into math, but this can probably be cast as a linear program and solved using something off-the-shelf like `scipy.optimize.linprog`
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html

eta/core/datasets.py
tylerganter: haha, yeah it's just a do-while loop. I changed the code, but the main condition is still used with a `if <this>: break`

eta/core/datasets.py
tylerganter: This is a very simple and bad solution which can easily be exchanged with `_greedy` but is not being used right now. The purpose was that multiple algorithms can be easily swapped in and out here and this one, while bad, is easy to understand and stands as an example.

10K is totally arbitrary, and also, I believe this is O(NM) where: N - number of samples, M - number of attribute values and it should be safe to assume that in most cases N >> M.

eta/core/datasets.py
```diff
@@ -1338,3 +1373,839 @@ def required(cls):
...
+        metadata = etav.VideoMetadata.build_for(self.data_path)
+        self.total_frame_count = metadata.total_frame_count
+        self.duration = metadata.duration
+        self.clip_end_frame = metadata.total_frame_count
+
+    def build(self, dir_path, filename, pretty_print=False):
+        '''Build the BuilderVideoRecord.
+
+        Args:
+            dir_path (str): path to write the files
```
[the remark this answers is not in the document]
benjaminpkane: Thanks. Good to know.

eta/core/datasets.py
```diff
@@ -1062,6 +1093,8 @@ class LabeledVideoDataset(LabeledDataset):
     which points to the `manifest.json` file for the dataset.
     '''
 
+    _BUILDER_CLS_FIELD = "eta.core.datasets.LabeledVideoDatasetBuilder"
```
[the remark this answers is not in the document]
benjaminpkane: Updates made!

There was a circularity issue which is why strings were originally used. That is all gone now, and the classes are generated via naming conventions now.

benjaminpkane:
@brimoor Here is some example code.

```
import eta.core.datasets as etad

d = etad.LabeledVideoDataset("/tmp/video_dataset/manifest.json")
builder = d.builder()
clipper = etad.Clipper(500, 500, 100)  # unit is frames
builder.add_transform(clipper)
balancer = etad.Balancer("attr_name", object_label="object_label", target_count=10)
builder.add_transform(balancer)
# and so on
builder.build("/tmp/out/manifest.json")
```

Ready for re-review!

eta/core/datasets.py
```diff
@@ -1338,3 +1375,842 @@ def required(cls):
...
+        keep_idxs = self._get_keep_idxs(occurrence_matrix, counts, target_count)
+
+        # STEP 4: modify the list of records
+        old_records = src.records
+        src.clear()
+        for ki in keep_idxs:
+            src.add(old_records[record_idxs[ki]])
+
+    def _get_occurrence_matrix(self, records):
+        '''Compute occurrence of each attribute value for each class
```
brimoor: FYI- our style doesn't require writing docstrings for private methods, if you don't want to

eta/core/datasets.py
```diff
@@ -1338,3 +1375,842 @@ def required(cls):
...
+        Returns:
+            a list of integer indices to keep
+        '''
+        b = counts - target_count
+
+        x = self._greedy(A, b)
+
+        return np.where(x == 0)[0]
+
+    def _random(self, A, b):
```
brimoor: @tylerganter what about refactoring these functions into a configurable argument of the `Balancer` constructor? You could support a `method="random" or "greedy"` argument that allows the user to choose which of these methods they want to use. Or, you could allow them to provide a `method_fcn` that implements their own `(A, b, score_fcn) => x` method

eta/core/datasets.py
```diff
@@ -1338,3 +1375,842 @@ def required(cls):
...
+            if cur_scores[i] >= best_score:
+                break
+
+            best_score = cur_scores[i]
+            best_x = x_matrix[:, i]
+            w = np.where(best_x == 0)[0]
+
+        return best_x
+
+    def _solution_score(self, vector):
```
brimoor: same comment about supporting customization of this function, and refactoring this implementation into a top-level function to avoid polluting the definition of this class

eta/core/datasets.py
brimoor: ^this would allow these functions to be defined as top-level functions, not as private methods of this class, which is cleaner I think (they are interchangeable bits, not intimately tied to the class)

brimoor: As is, this is a private method that is never used, which is basically like including commented code in a PR (discouraged). Upgrading to a top-level method makes it publicly available, so it is now constructive code :)

tylerganter: I made the algorithm a configurable in the initialization. I feel the algorithm is too inherently tied to balancing for me to want to expose them as non-private methods.

@brimoor the intention, which is something that may be a discussion point in regards to ETA convention, is to make it easy to reference if someone wanted to replace `greedy` with something else down the road. I don't think you would ever choose `random` over `greedy`, but it is a very simple algorithm that is in turn easy to understand and reference.

eta/core/datasets.py
```diff
@@ -478,14 +484,22 @@ def iter_labels(self):
...
-        '''Iterates over the paths lables files in the dataset.
+        '''Iterates over the paths labels files in the dataset.
 
         Returns:
             iterator: iterator over paths to labels files
         '''
         for record in self.dataset_index:
             yield os.path.join(self.data_dir, record.labels)
 
+    def iter_paths(self):
```
mattphotonman: This is useful, thanks for adding this!

PR #261, opened 2019-07-15
Sampler impl

MikeJeffers:


eta/core/datasets.py
```diff
@@ -1265,14 +1265,11 @@ class DatasetTransformer(object):
     Basically, strategy pattern
     '''
 
-    def __init__(self):
```
tylerganter: why get rid of this?

MikeJeffers: Because the sub-classes were calling this with `super` - which would throw the error.

tylerganter: So there's no reason for the sub-classes to call super. I would suggest that we remove those calls as well. And the `NotImplementedError` should be in there if it is **necessary** for the sub-classes to explicitly define `__init__`, which i don't think is the case, so I'm fine with deleting this

MikeJeffers: Right, I don't think there is yet any shared state to DatasetTransformers to be reasonably set in an __init__
Removed from other stubbed classes.

PR #264, opened 2019-07-15
Delete files in folder and get count of files

rpinnaka:
- To delete all the files in GoogleDrive folder and retain the empty folder

- To get the count of the number of files present in the GoogleDrive folder

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -639,6 +639,9 @@ class GoogleDriveStorageClient(StorageClient, NeedsGoogleCredentials):
                     logger.info(
                         "Failed to delete file '%s' from '%s' (%s)", filename, folder_id,
                         t.elapsed_time_str)
+        
+        if not len(self.list_files_in_folder(folder_id)):
+            logger.info("All files deleted in %s", folder_id)
 
     def count_files_in_folder(self, folder_id):
         '''Returns count of number of files in the Google Drive folder
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -15,6 +15,7 @@ Copyright 2018-2019, Voxel51, Inc.
 voxel51.com
 
 Brian Moore, brian@voxel51.com
+Ravali Pinnaka, ravali@voxel51.com
 '''
 # pragma pylint: disable=redefined-builtin
 # pragma pylint: disable=unused-wildcard-import
@@ -612,6 +613,32 @@ class GoogleDriveStorageClient(StorageClient, NeedsGoogleCredentials):
         '''
         self._service.files().delete(
             fileId=file_or_folder_id, supportsTeamDrives=True).execute()
+
+    def delete_all_files_in_folder(self, folder_id):
+        '''Deletes all the files in the Google Drive folder and retains the 
+        original folder.
+
+        Args:
+            folder_id: the ID of the Google Drive folder from which the files 
+                are to be deleted
+        
+        Returns:
+            an empty folder with no files
+        '''
+        files = self.list_files_in_folder(folder_id)
+        for f in files:
+            filename = f["name"]
+            try:
+                with etau.Timer() as t:
+                    self.delete(f["id"])
+                    logger.info(
+                    "File '%s' deleted from '%s' (%s)", filename, folder_id,
+                    t.elapsed_time_str)
+            except Exception as e:
+                if not skip_failures:
+                    raise GoogleDriveStorageClientError(e)
+
+        logger.info("All files deleted in %s", folder_id)
 
     def get_team_drive_id(self, name):
         '''Get the ID of the Team Drive with the given name.
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -640,7 +640,7 @@ class GoogleDriveStorageClient(StorageClient, NeedsGoogleCredentials):
                     "Failed to delete file '%s' from '%s' (%s)", filename,
                     folder_id, t.elapsed_time_str)
 
-        if not len(self.list_files_in_folder(folder_id)):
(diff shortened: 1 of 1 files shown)


eta/core/storage.py
```diff
@@ -613,6 +614,47 @@ def delete(self, file_or_folder_id):
...
+                    logger.info(
+                        "File '%s' deleted from '%s' (%s)", filename, folder_id,
+                        t.elapsed_time_str)
+            except Exception as e:
+                if not skip_failures:
+                    raise GoogleDriveStorageClientError(e)
+
+        logger.info("All files deleted in %s", folder_id)
+
+    def get_count_of_files_in_folder(self, folder_id):
```
brimoor: this could be implemented in one line as
```py
return len(self. list_files_in_folder(folder_id))
```

I'm not sure this is complex enough to warrant its own method, but if it stays then please:
- rename to `count_files_in_folder()` for brevity
- implement as one-liner above

eta/core/storage.py
```diff
@@ -613,6 +614,47 @@ def delete(self, file_or_folder_id):
...
+                with etau.Timer() as t:
+                    self.delete(f["id"])
+                    logger.info(
+                        "File '%s' deleted from '%s' (%s)", filename, folder_id,
+                        t.elapsed_time_str)
+            except Exception as e:
+                if not skip_failures:
+                    raise GoogleDriveStorageClientError(e)
+
+        logger.info("All files deleted in %s", folder_id)
```
brimoor: Not necessarily. What if a deletion failed and you skipped it?

eta/core/storage.py
```diff
@@ -613,6 +614,47 @@ def delete(self, file_or_folder_id):
...
+        for f in files:
+            filename = f["name"]
+            try:
+                with etau.Timer() as t:
+                    self.delete(f["id"])
+                    logger.info(
+                        "File '%s' deleted from '%s' (%s)", filename, folder_id,
+                        t.elapsed_time_str)
+            except Exception as e:
+                if not skip_failures:
```
brimoor: `skip_failures` is not defined. you copied this from another method but didn't implement this flag

eta/core/storage.py
```diff
@@ -613,6 +614,47 @@ def delete(self, file_or_folder_id):
         self._service.files().delete(
             fileId=file_or_folder_id, supportsTeamDrives=True).execute()
 
+    def delete_all_files_in_folder(self, folder_id):
```
brimoor: Why would someone want to delete all files in the folder but not the folder? Because you don't want the ID of the parent folder to change if you just deleted the entire folder with `delete()` and then recreated the empty folder?

This latter approach is likely way faster for deleting folders with lots of files in them.

eta/core/storage.py
```diff
@@ -613,6 +614,47 @@ def delete(self, file_or_folder_id):
...
+        '''Deletes all the files in the Google Drive folder and retains the
+        original folder.
+
+        Args:
+            folder_id: the ID of the Google Drive folder from which the files
+                are to be deleted
+
+        Returns:
```
mattphotonman: The function doesn't actually return anything.  In such as case, you can omit the "Returns" section of the doc string.

eta/core/storage.py
rpinnaka: wanted to have a method to get the count of files (for example, to get the count of validation/training images in the GoogleDrive folder). will implement in the one-liner mentioned, thanks.

eta/core/storage.py
rpinnaka: yes, will add it. thanks.

eta/core/storage.py
rpinnaka: will implement the flag, thanks.

eta/core/storage.py
rpinnaka: Some scenarios have been encountered where the old files were to be replaced with the new ones. Here, I wanted to remove only the files and retain the original folder to upload the new files. (doesn't want the ID of the parent folder to be changed, as this folder ID might have been mentioned elsewhere in the data source)

eta/core/storage.py
rpinnaka: removed it, thanks.

eta/core/storage.py
```diff
@@ -613,6 +614,46 @@ def delete(self, file_or_folder_id):
...
+            try:
+                with etau.Timer() as t:
+                    self.delete(f["id"])
+                    logger.info(
+                        "File '%s' deleted from '%s' (%s)", filename,
+                        folder_id, t.elapsed_time_str)
+            except Exception as e:
+                if not skip_failures:
+                    raise GoogleDriveStorageClientError(e)
+                    logger.info(
```
mattphotonman: I'm not sure if this line would ever be executed since it is after a `raise` statement.

eta/core/storage.py
```diff
@@ -613,6 +614,46 @@ def delete(self, file_or_folder_id):
...
+                        "Failed to delete file '%s' from '%s' (%s)", filename,
+                        folder_id, t.elapsed_time_str)
+
+        if not len(self.list_files_in_folder(folder_id)):
+            logger.info("All files deleted in %s", folder_id)
+
+    def count_files_in_folder(self, folder_id):
```
mattphotonman: Do you need to have this method now?  It doesn't look like you're actually using it.

eta/core/storage.py
rpinnaka: fixed it, thanks.

eta/core/storage.py
rpinnaka: it would be good to have a count method, but not mandatory though. It can removed, if not necessary.

PR #265, opened 2019-07-16
Clipper backend

kevinqi34:
Should we not allow users to clip to 1 frame. 
I can prevent them in the front end.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1241,7 +1241,7 @@ class BuilderVideoRecord(BuilderDataRecord):
             "clip_start_frame",
             "clip_end_frame",
             "duration",
-            "total_frame_count"
+            "total_frame_count",
         ]
         return attrs
 
@@ -1390,9 +1390,18 @@ class Clipper(DatasetTransformer):
                     clip_duration = int(end_frame - start_frame + 1)
                     if clip_duration < int(self.min_clip_len):
                         break
-                print(str(start_frame) + ' ' + str(end_frame))
+                self.add_clipping(start_frame, end_frame, record, src.records)
                 start_frame += self.clip_len
                 start_frame += self.stride
+
+    def add_clipping(self, start_frame, end_frame, old_record, records):
+        frame_rate = old_record.frame_rate
+        new_record = old_record.copy()
+        new_record.start_frame = start_frame
+        new_record.clip_end_frame = end_frame
+        new_record.total_frame_count = end_frame - start_frame + 1
+        new_record.duration = round((end_frame - start_frame) / frame_rate, 3)
+        records.append(new_record)
 
 
 class DatasetTransformerError(Exception):
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1195,6 +1195,7 @@ class BuilderVideoRecord(BuilderDataRecord):
         self.total_frame_count = metadata.total_frame_count
         self.duration = metadata.duration
         self.clip_end_frame = metadata.total_frame_count
+        self.frame_rate = metadata.frame_rate
 
     def build(self, dir_path, filename, pretty_print=False):
         self._extract_video_labels()
@@ -1348,9 +1349,17 @@ class SchemaFilter(DatasetTransformer):
 class Clipper(DatasetTransformer):
     '''
     Clip longer videos into shorter ones, and sample at some stride step
+
     '''
 
     def __init__(self, clip_len, stride, min_clip_len):
+        '''Creates a Clipper instance.
+
+        Args:
+            clip_len: number of frames per clip
+            stride: gap between clips in frames
+            min_clip_len: minimum number of frames allowed
+        '''
         self.clip_len = clip_len
         self.stride = stride
         self.min_clip_len = min_clip_len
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1372,7 +1381,27 @@ def transform(self, src):
...
+                self.add_clipping(start_frame, end_frame, record, src.records)
+                start_frame += self.clip_len
+                start_frame += self.stride
+
+    def add_clipping(self, start_frame, end_frame, old_record, records):
+        frame_rate = old_record.frame_rate
+        new_record = old_record.copy()
+        new_record.start_frame = start_frame
+        new_record.clip_end_frame = end_frame
+        new_record.total_frame_count = end_frame - start_frame + 1
```
benjaminpkane: `total_frame_count` and `duration` are with respect to the raw video. So you should not change these.

eta/core/datasets.py
```diff
@@ -1372,7 +1381,27 @@ def transform(self, src):
...
+                if end_frame > record.clip_end_frame:
+                    end_frame = record.clip_end_frame
+                    clip_duration = int(end_frame - start_frame + 1)
+                    if clip_duration < int(self.min_clip_len):
+                        break
+                self.add_clipping(start_frame, end_frame, record, src.records)
+                start_frame += self.clip_len
+                start_frame += self.stride
+
+    def add_clipping(self, start_frame, end_frame, old_record, records):
```
benjaminpkane: This can probably be a private function. Only `transform` will ever be called on a `DatasetTransformer`.

eta/core/datasets.py
```diff
@@ -1372,7 +1381,27 @@ def transform(self, src):
...
+        frame_rate = old_record.frame_rate
+        new_record = old_record.copy()
+        new_record.start_frame = start_frame
+        new_record.clip_end_frame = end_frame
+        new_record.total_frame_count = end_frame - start_frame + 1
+        new_record.duration = round((end_frame - start_frame) / frame_rate, 3)
```
benjaminpkane: Could you remove `frame_rate`? Knowing that you don't have to change `duration`, you shouldn't need it.

eta/core/datasets.py
```diff
@@ -1372,7 +1381,27 @@ def transform(self, src):
...
+                    clip_duration = int(end_frame - start_frame + 1)
+                    if clip_duration < int(self.min_clip_len):
+                        break
+                self.add_clipping(start_frame, end_frame, record, src.records)
+                start_frame += self.clip_len
+                start_frame += self.stride
+
+    def add_clipping(self, start_frame, end_frame, old_record, records):
+        frame_rate = old_record.frame_rate
+        new_record = old_record.copy()
```
benjaminpkane: Thank you!

benjaminpkane:
> Should we not allow users to clip to 1 frame.
> I can prevent them in the front end.

I would say it's not necessary. If that is what they set as their minimum, then so be it.

eta/core/datasets.py
kevinqi34: oh

eta/core/datasets.py
kevinqi34: > > Should we not allow users to clip to 1 frame.
> > I can prevent them in the front end.
> 
> I would say it's not necessary. If that is what they set as their minimum, then so be it.

but it seems like videos of 1 frame don't play

eta/core/datasets.py
kevinqi34: thanks for clarifying

eta/core/datasets.py
benjaminpkane: Okay, if you think it will avoid problems like that, then sure. I don't have a strong opinion either way.

kevinqi34:
There seems to be a little descrepancy between the inputs and the actual video outputted. 
For instance, I was trying to clip the example down to only 3 frames but after going through the process of converting frames to seconds during the ffmpeg call, it seems that the clip was lengthened by a few frames, likely due to a decimal rounding. Do you guys forsee that this will be an issue?

benjaminpkane:
> There seems to be a little descrepancy between the inputs and the actual video outputted.
> For instance, I was trying to clip the example down to only 3 frames but after going through the process of converting frames to seconds during the ffmpeg call, it seems that the clip was lengthened by a few frames, likely due to a decimal rounding. Do you guys forsee that this will be an issue?

Yes, we should talk about this.

kevinqi34:
> > There seems to be a little descrepancy between the inputs and the actual video outputted.
> > For instance, I was trying to clip the example down to only 3 frames but after going through the process of converting frames to seconds during the ffmpeg call, it seems that the clip was lengthened by a few frames, likely due to a decimal rounding. Do you guys forsee that this will be an issue?
> 
> Yes, we should talk about this.

Actually my mistake, it was a syntax bug that was making it get clips from frame 1 always.

kevinqi34:
It's actually really accurate.

eta/core/datasets.py
```diff
@@ -1372,7 +1380,24 @@ def transform(self, src):
...
+            start_frame = record.clip_start_frame
+            while start_frame <= record.clip_end_frame:
+                end_frame = start_frame + self.clip_len - 1
+                if end_frame > record.clip_end_frame:
+                    end_frame = record.clip_end_frame
+                    clip_duration = int(end_frame - start_frame + 1)
+                    if clip_duration < int(self.min_clip_len):
+                        break
+                self._add_clipping(start_frame, end_frame, record, src.records)
+                start_frame += self.clip_len
```
benjaminpkane: Stride is a bit of a misnomer if used in this way. If we remove this line though, it makes sense.

kevinqi34: what do you mean?

PR #268, opened 2019-07-16
Balancer

tylerganter:
Balancer up and running! I still need to do some more testing. And this certainly isn't the "optimal solution" so much as something that seemed like a reasonable balance of algorithm complexity, compute efficiency and result accuracy.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1338,10 +1338,33 @@ class Balancer(DatasetTransformer):
 
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         '''
+
         old_records = src.records
         src.clear()
 
         # STEP 1: Get attribute value(s) for every record
+        helper_list = self._to_helper_list(old_records)
+
+        # STEP 2: balance the records
+        counts, to_remove = self._get_counts_and_to_remove(helper_list)
+
+        for x in counts:
+            print(x)
+        print('~~~~~~~~~~~~~~~~~~')
+        for x in to_remove:
+            print(x)
+
+        # def entropy(pk):
+        #     import numpy as np
+        #     pk = np.array(pk)
+        #     pk = pk / np.sum(pk)
+        #     S = -np.sum(pk * np.log(pk + 1e-9), axis=0)
+        #     return S / -np.log(1 / pk.shape[0])
+        # print(entropy(counts.values()))
+
+        # src.add(old_records[0])
+
+    def _to_helper_list(self, old_records):
         helper_list = []
 
         for i, record in enumerate(old_records):
@@ -1364,8 +1387,9 @@ class Balancer(DatasetTransformer):
                         helper_list.append((i, [attr.value]))
                         break
 
-        # STEP 2: balance the records
-        # @TODO CONTINUE HERE
+        return helper_list
+
+    def _get_counts_and_to_remove(self, helper_list, median_point=0.25):
         counts = {}
         for idx, attr_values in helper_list:
             for attr_value in attr_values:
@@ -1374,10 +1398,16 @@ class Balancer(DatasetTransformer):
                 else:
                     counts[attr_value] += 1
 
-        for key, value in counts.items():
-            print('{} - {}'.format(key, value))
+        counts = list(counts.items())
+        counts.sort(key=lambda x: x[1])
 
-        # src.add(old_records[0])
+        target_count = counts[int(len(counts) * median_point)][1]
+
+        to_remove = [(k, v - target_count) for k, v in counts]
+        # @TODO leave the negatives in? (comment this line out)
+        to_remove = [(k, max(v, 0)) for k, v in to_remove]
+
+        return counts, to_remove
 
 
 class SchemaFilter(DatasetTransformer):
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1347,25 +1347,11 @@ class Balancer(DatasetTransformer):
 
     def transform(self, src):
         '''
+        Modify the BuilderDataset records by removing records until the
+        target attribute is ~roughly~ balanced for each value.
 
         Args:
             src (BuilderDataset): the dataset builder
-
-        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-        categorical_attribute = image_labels.attrs[0]
-        categorical_attribute.name
-        categorical_attribute.value
-
-        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-        detected_object = image_labels.objects[0]
-        detected_object.label
-        categorical_attribute = detected_object.attrs[0]
-        categorical_attribute.name
-        categorical_attribute.value
-
-        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         '''
 
         # STEP 1: Get attribute value(s) for every record
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -28,6 +28,7 @@ import six
 import logging
 import os
 import random
+from collections import Counter, defaultdict
 
 import numpy as np
 
@@ -1546,7 +1547,7 @@ class Balancer(DatasetTransformer):
         for i, record in enumerate(records):
             labels = record.get_labels()
             NO_ID = 'NO_ID'
-            helper_dict = {}
+            helper_dict = defaultdict(set)
 
             for frame_no, frame in labels.frames.items():
                 if (frame_no < record.clip_start_frame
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1370,19 +1373,294 @@ def transform(self, src):
...
+        keep_idxs = self._get_keep_idxs(target_remove, helper_list)
+
+        # STEP 4: modify the list of records
         old_records = src.records
         src.clear()
-        pass
+        for ki in keep_idxs:
+            src.add(old_records[helper_list[ki][0]])
 
+    def _to_helper_list(self, records):
```
kevinqi34: maybe split this into two functions. 
helper_video
helper_image 

And call them from _to_helper_list.
This function is rather long.

eta/core/datasets.py
```diff
@@ -1370,19 +1373,294 @@ def transform(self, src):
...
+        for idx, attr_values in helper_list:
+            for attr_value in attr_values:
+                if attr_value not in counts:
+                    counts[attr_value] = 1
+                else:
+                    counts[attr_value] += 1
+
+        # sort by count
+        counts_list = list(counts.items())
+        counts_list.sort(key=lambda x: x[1])
```
kevinqi34: I don't know if this is going to be an issue, but the O(nlogn) sort operation here could be mitigated by
having the list sort on construction.

tylerganter: this is sorting the attribute values, so max 100 values in a crazy large case

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+                            balanced, one per unique object, if using objects.
+                            For example: ['red', 'red', 'green'] would imply
+                            three objects with the 'color' attribute in this
+                            record.
+        '''
+
+        if not len(records):
+            return []
+
+        elif type(records[0]) == BuilderImageRecord:
```
mattphotonman: use `isinstance()` instead of `type()`

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+            if self.object_label:
+                return self._to_helper_list_image_objects(records)
+            else:
+                return self._to_helper_list_image(records)
+
+        elif type(records[0]) == BuilderVideoRecord:
```
mattphotonman: use `isinstance()` instead of `type()`

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+
+                    for attr in detected_object.attrs:
+                        if attr.name == self.attr_name:
+                            obj_idx = (
+                                detected_object.index
+                                if detected_object.index is not None
+                                else NO_ID
+                            )
+
+                            if obj_idx in helper_dict:
```
mattphotonman: You can use a `defaultdict(set)` to avoid this if statement here.

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+        return helper_list
+
+    def _to_helper_list_video_objects(self, records):
+        '''Balancer._to_helper_list for object attributes in videos'''
+        helper_list = []
+
+        for i, record in enumerate(records):
+            labels = record.get_labels()
+            NO_ID = 'NO_ID'
+            helper_dict = {}
```
mattphotonman: `helper_dict = defaultdict(set)` (see below)

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+                            else:
+                                helper_dict[obj_idx] = set([attr.value])
+
+                            break
+
+            # At this point, the keys of helper dict are unique
+            # object indices for objects of type self.object_label.
+            # The values are unique attribute values for self.attr_name.
+
+            if len(helper_dict):
```
mattphotonman: `if helper_dict:`

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+            "Unknown record type: {}".format(type(records[0])))
+
+    def _to_helper_list_image(self, records):
+        '''Balancer._to_helper_list for image attributes'''
+        helper_list = []
+
+        for i, record in enumerate(records):
+            labels = record.get_labels()
+
+            for attr in labels.attrs:
```
mattphotonman: Maybe we should be keeping only `CategoricalAttribute`s and `BooleanAttribute`s here, (i.e. not `NumericAttribute`s).

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+                            break
+
+            # At this point, the keys of helper dict are unique
+            # object indices for objects of type self.object_label.
+            # The values are unique attribute values for self.attr_name.
+
+            if len(helper_dict):
+                helper = (i, [])
+                for s in helper_dict.values():
+                    helper[1].extend(list(s))
```
mattphotonman: `helper[1].extend(s)` would work fine without the list conversion.

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+                helper_list.append(helper)
+
+        return helper_list
+
+    def _get_target_remove(self, helper_list):
+        '''
+
```
mattphotonman: Style: add one sentence description, then
```python
Args:
    helper_list: ...

Returns:
    ...
```

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+        helper_list: return value from Balancer._to_helper_list(...)
+        return: a dictionary with:
+            keys: the values of the attribute to balance
+            values: the target count to remove of this attribute value
+        '''
+        # compute counts for each attribute value
+        counts = {}
```
mattphotonman: `counts = Counter()` (see below)

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+        for idx, attr_values in helper_list:
+            for attr_value in attr_values:
+                if attr_value not in counts:
```
mattphotonman: You can avoid this if statement if you use a `Counter`.

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+                    counts[attr_value] = 1
+                else:
+                    counts[attr_value] += 1
+
+        # sort by count
+        counts_list = list(counts.items())
+        counts_list.sort(key=lambda x: x[1])
+
+        # integer target value
+        target_count = counts_list[int(len(counts_list) * self.median_point)][1]
```
mattphotonman: You can use `numpy.quantile` ([link](https://docs.scipy.org/doc/numpy/reference/generated/numpy.quantile.html)), if you first put the values into a numpy array.  This will save you having to worry about things like ties.

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+
+        target_remove = [(k, v - target_count) for k, v in counts_list]
```
mattphotonman: Can make into a dictionary in one line with a dictionary comprehension:
```python
target_remove = {k: v - target_count for k, v in counts_list}
```

eta/core/datasets.py
```diff
@@ -1372,18 +1375,336 @@ def transform(self, src):
...
+
+        return target_remove
+
+    def _get_keep_idxs(self, target_remove, helper_list):
+        '''
+        Algorithm fun! This function chooses the set of records to keep (and
+        remove).
+
+        There's still plenty of potential for testing and improvement here.
+
```
mattphotonman: `Args:`

eta/core/datasets.py
```diff
@@ -1370,19 +1373,294 @@ def transform(self, src):
...
+    def __init__(self, attribute_name, object_label=None,
+                 median_point=0.25, negative_power=5.0):
+        '''
+
+        Args:
+            attribute_name (str): the name of the attribute to balance by
+            object_label (str): the name of the object label that the
+                attribute_name must be nested under. If this is None, it is
+                assumed that the attributes are Image/Frame level attrs.
+            median_point (float): value between [0, 1] to specify what the
```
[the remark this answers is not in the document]
tylerganter: woah new vocab for me!

eta/core/datasets.py
```diff
@@ -1370,19 +1373,294 @@ def transform(self, src):
...
+            record_id: integer ID of the corresponding old_record
+            list_of_values: list of attribute values for the attribute to be
+                            balanced, one per unique object, if using objects.
+                            For example: ['red', 'red', 'green'] would imply
+                            three objects with the 'color' attribute in this
+                            record.
+        '''
+        helper_list = []
+
+        if not len(records):
```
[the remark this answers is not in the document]
tylerganter: this is a nit point for me. I think that the way I wrote it helps emphasize what my intention was. If this is a strict ETA rule, then I'll change it, but otherwise I think that adding `len(...)` makes it clear that I am not check for `None` or `False` or some other truthy term.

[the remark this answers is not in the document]
mattphotonman: Whether it's a strict ETA rule is up to all of us!  I think it is mentioned in the style guide somewhere, but I don't think this way is so bad.  What I would object to more is something like `if len(records) == 0:`.

eta/core/datasets.py
tylerganter: same as rebuttal above

eta/core/datasets.py
tylerganter: I've been assuming categorical...but would there be an issue for it to be boolean or numeric? I don't see the problem. unless perhaps numeric is a continuous space...

eta/core/datasets.py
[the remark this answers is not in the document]
tylerganter: Agreed, to me, `len(records)` without `== 0` is both concise and clear. So I'm going to resolve this.

PR #270, opened 2019-07-19
Clipper Transform needs some arg bounding

MikeJeffers:
Some params could be provided that could cause infinite loops and other bad things.
An exception is raised to prevent this.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1841,9 +1841,14 @@ class Clipper(DatasetTransformer):
             stride_len: stride (step size)
             min_clip_len: minimum number of frames allowed
         '''
-        self.clip_len = clip_len
-        self.stride_len = stride_len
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1837,13 +1837,19 @@ def __init__(self, clip_len, stride_len, min_clip_len):
...
+            min_clip_len (int): minimum number of frames allowed, must be > 0
+                                and greater than clip_len
         '''
-        self.clip_len = clip_len
-        self.stride_len = stride_len
-        self.min_clip_len = min_clip_len
+        self.clip_len = int(clip_len)
+        self.stride_len = int(stride_len)
+        self.min_clip_len = int(min_clip_len)
+        bad_args = self.clip_len < 1 or self.stride_len < 1
```
kevinqi34: cannot believe I forgot this.

PR #275, opened 2019-07-25
BigContainer

benjaminpkane:
I need to test this still some, but if anyone has any thoughts about the design in general, I'd love to hear them.

Right now `Big*SetLabels` subclass `*SetLabels` and just override `from_dict` and `__init__`. I intend to change that so both classes are defined concretely from a single parent class.

I'm more curious if the`DirectoryContainer` class that was written is the right approach. It seemed like the path of least resistance.

Naming can change as well.


eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        return (self._load_ele(path) for path in self._paths)
+
+    def __getitem__(self, idx):
+        return self._load_ele(self._ele_path(idx))
+
+    def __setitem__(self, idx, ele):
+        ele.write_json(self._ele_path(idx))
+
+    def __delitem__(self, idx):
+        etau.delete_file(self._ele_path(idx))
```
mattphotonman: Do we want delete to actually delete the file?  I could see cases where we would not want to include the item in the list anymore, but not actually want to delete the underlying file.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+
+        Args:
+            filters: a list of functions that accept elements and return
+                True/False
+            match: a function (usually `any` or `all`) that accepts an iterable
+                and returns True/False. Used to aggregate the outputs of each
+                filter to decide whether a match has occurred. The default is
+                `any`
+        '''
+        self._filter_elements(filters, match)
```
mattphotonman: I'm confused: it looks like this doesn't actually remove elements, just returns those that match the filter.  Am I missing something?

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        setattr(container, self._ELE_DIR_ATTR, new_ele_dir)
+        for idx, ele in self:
+            if idx in inds:
+                container.add(ele)
+        return container
+
+    def clear(self):
+        '''Deletes all elements from the container.'''
+        super(DirectoryContainer, self).clear()
+        etau.delete_dir(self.__elements_dir__)
```
mattphotonman: Hmm, what if you have other stuff in that directory?

eta/core/video.py
```diff
@@ -1056,6 +1056,70 @@ def from_dict(cls, d):
...
+        Args:
+            new_images_dir (str): directory path to store the copy's data
+
+        Returns:
+            BigVideoSetLabels
+        '''
+        new_set = copy.deepcopy(self)
+        new_set.videos = self.videos.copy(new_videos_dir)
+        return new_set
+
```
mattphotonman: Will you have to override `VideoSetLabels.sort_by_filename()`?  Will `sorted()` do what you want it to do on a 'DirectoryContainer`?

mattphotonman:
Left some one-off comments.  I might need to take another look as I don't know if I was following all of the setting and deleting logic.  The `DirectoryContainer` approach looks good overall.  My one major question would be whether we want to force everything to be in the same directory, or just want to view it as a collection of paths that could be in different directories.  (The latter approach is similar to how `DataRecords` is used.) Note that some `DirectoryContainer` methods, like `clear()`, also assume that you're not really putting anything else in that directory, or you risk it getting deleted as a side-effect.

eta/core/video.py
benjaminpkane: `sorted()` seems to work. I'll be doing more testing today, of course.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        setattr(self, self._ELE_DIR_ATTR, kwargs[self._ELE_DIR_ATTR])
+
+        setattr(
+            self,
+            self._ELE_ATTR,
+            kwargs.get(
+                self._ELE_ATTR,
+                [
+                    os.path.splitext(os.path.basename(i))[0]
+                    for i in etau.multiglob(".json",
```
brimoor: Just use `etau.list_files` here, now that `backing_dir` only contains elements

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+    '''
+
+    #
+    # The name of the attribute that will store the path to the json data files
+    #
+    # Subclasses MUST set this field
+    #
+    _ELE_DIR_ATTR = None
+
+    def __init__(self, **kwargs):
```
brimoor: I really like how the "JSON serializable part of this is just a lightweight list of uuids. Nice idea!

For `Big<Class>` objects, I think we should support `from_zip()` and `to_zip()` methods as well, so that one can easily get ahold of all the data.

To implement `to_zip("/path/to/output.zip")`, I would propose that you arrange the files in a tempdir with structure:
```py
dir/
    index.json
    <elements>/
        <uuid>.json
```
then zip it and delete the tempdir.

eta/core/video.py
mattphotonman: I guess what I meant was that `sorted()` would return a `list` and not a `DirectoryContainer`.  Won't that cause problems for future operations?  (It violates the contract that `BigVideoSetLabels.videos` has to be a `VideoSetLabelsDirectory`.)

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
 
-        return cls(**{
-            cls._ELE_ATTR: [ele_cls.from_dict(dd) for dd in d[cls._ELE_ATTR]]
-        })
+
+class DirectoryContainer(Container):
+    '''Container that represents a list of serializable objects in a
+    directory on disk. Any accessing of an element in the list causes a READ
+    from disk and updating any elements in the list REQUIRES setting the item
+    explicitly (and causes a WRITE to disk).
```
brimoor: I just remembered one more feature request: implement `to_container` and `from_container` that will convert between `Big<ContainerClass>` and `<ContainerClass>`. The convention would be that these classes would be defined in the same module. If `<ContainerClass>` does not exist, the user gets an error message.

`to_container()` would be the stylistic thing for users of these classes, not `in_memory_list = [e for e in c]`

eta/core/serial.py
mattphotonman: How about having that directory structure for the actual backing dir?  (Then the functionality would be similar to a `LabeledDataset`.)

eta/core/serial.py
benjaminpkane: Yes, realizing now that this function as well as `_filter_elements()` needs more work.

eta/core/serial.py
benjaminpkane: Will do.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
             # Validates the cls settings
             cls()
-            # Parse using provided class
-            ele_cls = cls._ELE_CLS
 
-        return cls(**{
-            cls._ELE_ATTR: [ele_cls.from_dict(dd) for dd in d[cls._ELE_ATTR]]
-        })
+
+class DirectoryContainer(Container):
```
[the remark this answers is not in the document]
benjaminpkane: Concise and meaningful! Will do.

eta/core/serial.py
mattphotonman: Yes, I now understand this better and agree.  The `backing_dir` should be solely managed by the `BigContainer` class, and should start out empty.  (This would be similar to `LabeledDataset`.) The general workflow would be that you create an empty `BigContainer` and then copy over initial files from somewhere else, e.g. a `LabeledDataset`. I think that you should add the following methods for a `BigContainer`:
- create a `BigContainer` from a list of existing paths to files.  This would ideally copy the files using disk operations to the `backing_dir`, rather then reading them into memory and writing them into the `backing_dir`.
- create a `BigVideoSetLabels` from a `LabeledVideoDataset`, which similarly copies over all of the labels files from the `LabeledVideoDataset`.  (Similarly for a `BigImageSetLabels`.)

eta/core/serial.py
brimoor: I thought about recommending that as well. Would be curious to hear Ben's thoughts. It is perhaps convenient to have independent control over the backing directory. Of course, that would be true for `LabeledDataset` as well...

eta/core/serial.py
brimoor: I would do this as `BigContainer.from_files()` and `VideoSetLabels.from_labeled_video_dataset()` and `BigVideoSetLabels.from_labeled_video_dataset()`

Same for `ImageSetLabels` and `LabeledImageDataset`

eta/core/image.py
```diff
@@ -458,6 +448,70 @@ def from_dict(cls, d):
...
+    _ELE_CLS = ImageLabels
+
+    _ELE_CLS_FIELD = "LABELS_CLS"
+
+    _ELE_ATTR = "labels_map"
+
+    _ELE_DIR_ATTR = "labels_dir"
+
+
+class BigImageSetLabels(ImageSetLabels):
```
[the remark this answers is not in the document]
benjaminpkane: Following up. That all makes sense to me! It would make everything much more manageable. Just don't know how much existing code assumes a simple `list` as the container in the `ImageSetLabels`/`VideoSetLabels`.

It seems worth the change probably, regardless.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+    _ELE_DIR_ATTR = None
+
+    def __init__(self, **kwargs):
+        '''Creates a DirectoryContainer instance.
+
+        NOTE: The <elements> uuid list is generated dynamically from the path if
+        it is not provided.
+
+        Args:
+            <elements_dir>: the directory path to store the data. This argument
```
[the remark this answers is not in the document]
benjaminpkane: Done!

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        '''
+        try:
+            etau.delete_dir(new_ele_dir)
+        except OSError:
+            pass
+        container = self.__class__(**{self._ELE_DIR_ATTR: new_ele_dir})
+        container.add_container(self)
+        return container
+
+    def append(self, ele):
```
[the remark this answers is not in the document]
benjaminpkane: `ImageSetLabels` and `VideoSetLabels` use append currently, that was the only reason I added append.

I'll circle back to this.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+    def _ele_dir(self):
+        return getattr(self, self._ELE_DIR_ATTR)
+
+    def _ele_subdir_pattern(self, idx):
+        return os.path.join(*list(self._ele_filename(idx)[:5]))
+
+    def _ele_filename(self, idx):
+        if idx > len(self):
+            raise ContainerError("Index %d out of bounds" % idx)
+        elif idx == len(self):
```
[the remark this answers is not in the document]
benjaminpkane: Good point. I've moved this logic to `add`.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        return "%s.json" % self.__elements__[idx]
+
+    def _ele_path(self, idx):
+        return os.path.join(
+            self._ele_dir,
+            self._ele_subdir_pattern(idx),
+            self._ele_filename(idx)
+        )
+
+    def _ele_path_by__uuid(self, _uuid):
```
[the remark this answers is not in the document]
benjaminpkane: Updates made!

eta/core/serial.py
benjaminpkane: I've updated the functionality, though I haven't been able to test it yet.

eta/core/serial.py
benjaminpkane: Updates made!

`from_paths`, `from_dir`, and `from_zip` are now present. I've tested some, but not enough.

eta/core/video.py
benjaminpkane: Okay, yes. I can override `sort_by_filename` for the `Big*` classes.

benjaminpkane:
I've made some updates. I probably missed a comment or two, and there is still a bit more work to be done, so I'll leave it as a draft for now.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+                filter to decide whether a match has occurred. The default is
+                `any`
+
+        Returns:
+            the number of elements in the container that match the filters
+        '''
+        return len(self._filter_elements(filters, match=match))
+
+    def get_matches(self, new_ele_dir, filters, match=any):
+        '''Gets elements matching the given filters.
```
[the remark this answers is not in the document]
benjaminpkane: Right. Docstring has been updated. `get_matches` uses `extract_inds` to return a new `BigContainer`.

eta/core/serial.py
benjaminpkane: As a `Serializable` object, my feeling is that having the "index" writable to any location using `write_json` is more consistent.

eta/core/serial.py
```diff
@@ -670,32 +705,325 @@ def from_dict(cls, d):
...
+        indices.
+
+        Args:
+            inds: a list of indices of the elements to keep
+
+        Returns:
+            DirectoryContainer
+        '''
+        inds = set(inds)
+        container = copy.deepcopy(self)
```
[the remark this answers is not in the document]
benjaminpkane: I believe this is fixed. It was definitely incorrect before, but now the elements is created properly for the new container.

mattphotonman:
> I've made some updates. I probably missed a comment or two, and there is still a bit more work to be done, so I'll leave it as a draft for now.

Let us know when you want us to take another look.  Looks like you upgraded this PR from a draft to "Open".

benjaminpkane:
> > I've made some updates. I probably missed a comment or two, and there is still a bit more work to be done, so I'll leave it as a draft for now.
> 
> Let us know when you want us to take another look. Looks like you upgraded this PR from a draft to "Open".

Yes, sorry, it's ready for review! I will have a `BigSet`, `BigImageSetLabels`, and `BigVideoSetLabels` up by tomorrow morning.

eta/core/serial.py
```diff
@@ -1124,17 +1113,477 @@ def _filter_elements(self, filters, match):
...
+        self._validate_cls()
+
+        self._backing_dir = None
+        self._set_backing_dir(backing_dir)
+
+        if self._ELE_ATTR in kwargs:
+            etau.ensure_dir(self.backing_dir)
+            elements = kwargs[self._ELE_ATTR]
+        else:
+            etau.ensure_empty_dir(self.backing_dir)
```
brimoor: Note that I removed the code that would try to automatically populate a BigContainer based on the contents of the backing directory. I did this because I think `BigContainer(backing_dir)` should throw an error if `backing_dir` is not empty.

The user can use `BigContainer.from_paths` or `BigContainer.from_dir` if they want to automatically populate a `BigContainer` from disk

brimoor:
Test code:

```py

# rm -rf /tmp/zzz*

import eta.core.geometry as etag

point = etag.LabeledPoint("origin", etag.RelativePoint(0, 0))

points = etag.BigLabeledPointContainer("/tmp/zzz")
points.add(point)

print(points[0])
print(points)

points.to_archive("/tmp/zzz.tar.gz")
points2 = etag.BigLabeledPointContainer.from_archive("/tmp/zzz.tar.gz", "/tmp/zzz2")
print(points2)

points2.move("/tmp/zzz2b")
print(points2)

points2.add(point)
points2.add_container(points)

points2.to_archive("/tmp/zzz2.zip")
points2c = etag.BigLabeledPointContainer.from_archive("/tmp/zzz2.zip", "/tmp/zzz2c")
print(points2c)

points2.delete_inds([0])
points2.keep_inds([1])
print(points2)

points2.add(point)
points2.add_container(points)

points3 = points2.extract_inds("/tmp/zzz3", [0, 1])

paths = list(points3._ele_paths)
points4 = etag.BigLabeledPointContainer.from_paths("/tmp/zzz4", paths)
print(points4)

points5 = etag.BigLabeledPointContainer.from_dir("/tmp/zzz5", "/tmp/zzz4")
print(points5)
```

benjaminpkane:
Thanks for the help!

PR #276, opened 2019-07-29
Adding a Set class and using it in ImageSetLabels and VideoSetLabels

brimoor:
This PR adds an `eta.core.serial.Set` class that provides O(1) key-based lookup into a homogeneous list of Serializable objects. This is the "set" analogue of the `eta.core.serial.Container` "list" class.

This PR also updates `ImageSetLabels` and `VideoSetLabels` to internally use `Set`. A few backwards incompatible changes were made:
- removed support for list-index-based access of the `*SetLabels` classes
- removed unused methods `has_*_labels_for_filename()` and `get_*_labels_for_filename()`

(Upon inspection, these deprecated features were not being used anywhere, so this is safe)


eta/core/serial.py
```diff
@@ -406,24 +407,424 @@ def _recurse(v, reflective):
...
+    def extract_keys(self, keys):
+        '''Creates a new set having only the elements with the given keys.
+
+        Args:
+            keys: an iterable of keys of the elements to keep
+
+        Returns:
+            a Set
+        '''
+        elements = copy.deepcopy(self._get_elements(keys))
```
benjaminpkane: So the assumption for `Container` and `Set` is that subclasses do not add there own attributes, i.e. `Container`s and `Set`s are containers and sets and nothing more?

Or is it just `extract_keys`, etc. do not copy those attributes over to returned instance...

benjaminpkane:
Also, are you still of the opinion that the forthcoming `BigSet` should subclass `BigContainer`? It seems somewhat unwieldy. `sort_by`, `add_container`, etc. would be present in `BigSet`, and `BigSet` would require changing `Container`'s `__elements__` from a `list` to an `OrderedDict` (I would think).

eta/core/serial.py
brimoor: Good point. I was trying to avoid copying the entire object, but that's a bit safer in case subclasses have other stuff in them. Fixed

brimoor:
@benjaminpkane I don't think `BigSet` should subclass `BigContainer`. I was just commenting that it could reuse a lot of the functionality

PR #277, opened 2019-07-30
Fixed bug in `__iter__()` method of Set class

mattphotonman:
`itervalues()` does not return an iterator.

### DIFF
--- a/eta/core/serial.py
+++ b/eta/core/serial.py
@@ -519,7 +519,8 @@ class Set(Serializable):
         return key in self.__elements__
 
     def __iter__(self):
-        return itervalues(self.__elements__)
+        for v in itervalues(self.__elements__):
+            yield v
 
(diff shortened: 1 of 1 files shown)


eta/core/serial.py
```diff
@@ -519,7 +519,8 @@ def __contains__(self, key):
         return key in self.__elements__
 
     def __iter__(self):
-        return itervalues(self.__elements__)
+        for v in itervalues(self.__elements__):
```
brimoor: hmm I guess it was bad of me to return an `iterable`, not actually an `iterator`...

You can write this more directly as:
```py
def __iter__(self):
    return iter(itervalues(self.__elements__))
```

mattphotonman: Oh sure, I thought that would look worse for some reason, but that works.

PR #279, opened 2019-07-30
BigSet and ImageSetLabels/VideoSetLabels refactor (converted to Set subclasses)

benjaminpkane:
Updated.

`BigSet` with optional temp storage functionality.

`ImageSetLabels` and `VideoSetLabels` are now direct subclasses of `Set`.

`BigImageSetLabels` and `BigVideoSetLabels` implementations, as well.

### DIFF
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -647,8 +647,35 @@ class ImageSetLabels(Set):
         Returns:
             an ImageSetLabels with the requested labels
         '''
-        return self.extract_keys(filenames)
+        # Intentionally bypass schema checking, for efficiency
+        schema = self.schema
+        self.schema = None
+        image_set_labels = self.extract_keys(filenames)
+        self.schema = schema
+        return image_set_labels
 
+    def get_matches(self, filters, match=any):
+        '''Returns an ImageSetLabels with labels matching the given filters.
+        The ImageLabels are passed by reference, not copied.
+
+        Args:
+            filters: a list of functions that accept elements and return
+                True/False
+            match: a function (usually `any` or `all`) that accepts an iterable
+                and returns True/False. Used to aggregate the outputs of each
+                filter to decide whether a match has occurred. The default is
+                `any`
+
+        Returns:
+            an ImageSetLabels with labels matching the filters
+        '''
+        # Intentionally bypass schema checking, for efficiency
+        schema = self.schema
+        self.schema = None
+        image_set_labels = super(ImageSetLabels, self).get_matches(
+            filters, match=match)
+        self.schema = schema
+        return image_set_labels
 
     def sort_by_filename(self, reverse=False):
         '''Sorts the ImageLabels in this instance by filename.
--- a/eta/core/serial.py
+++ b/eta/core/serial.py
@@ -813,7 +813,225 @@ class Set(Serializable):
         return etau.get_class(d["_CLS"])
 
 
-class BigSet(Set):
+class BigMixin(object):
+    '''Mixin class for BigContainer and BigSet.
+
+    `backing_dir` is an optional keyword argument for any methods that require
+    initializing a new instance of a Big object. When a `backing_dir` is not
+    provided, a tmp directory is used and is deleted upon the Big object's
+    destruction.
+
+    `move()` can be used to move the Big object between a set (persistent)
+    backing directory and a temporary backing directory.
+
+    See BigSet and BigContainer for concrete implementations.
+    '''
+
+    def __init__(self):
+        raise NotImplementedError("subclasses must implement __init__()")
+
+    def __del__(self):
+        if self._temp_storage:
+            etau.delete_dir(self.backing_dir)
+
+    @property
+    def backing_dir(self):
+        '''The backing directory for this Big object.'''
+        if self._backing_dir is not None:
+            return self._backing_dir
+        return self._tempdir
+
+    def clear(self):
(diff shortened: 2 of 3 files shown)


eta/core/video.py
```diff
@@ -1455,6 +1455,17 @@ def from_dict(cls, d):
         return cls(videos=videos, schema=schema)
 
 
+class _BigVideoLabelsSet(BigSet, _VideoLabelsSet):
+    '''Internal set type used by BigVideoSetLabels to store labels.'''
+
+
+class BigImageSetLabels(VideoSetLabels):
```
brimoor: BigVideoSetLabels

eta/core/image.py
```diff
@@ -786,6 +786,17 @@ def from_dict(cls, d):
         return cls(images=images, schema=schema)
 
 
+class _BigImageLabelsSet(BigSet, _ImageLabelsSet):
+    '''Internal set type used by BigImageSetLabels to store labels.'''
+
+
+class BigImageSetLabels(ImageSetLabels):
```
brimoor: All the code finally pays off! Pretty slick that this is all we have to write to get a `Big` version of the class!

eta/core/image.py
```diff
@@ -786,6 +786,17 @@ def from_dict(cls, d):
...
+    '''ImageSetLabels which uses a BigSet (_BigImageLabelsSet) to store labels.
```
brimoor: No need to mention `_BigImageLabelsSet` explicitly. Users don't need to read about private classes.

I'd explicitly mention that `BigImageSetLabels` behaves identically to `ImageSetLabels` except that each element is stored on disk.

brimoor:
@benjaminpkane feel free to introduce mixins if you see a good opportunity + value to do it. There are definitely opportunities. Even for `Container` and `Set`

eta/core/serial.py
```diff
@@ -797,6 +797,382 @@ def _validate_dict(cls, d):
...
+            filters: a list of functions that accept elements and return
+                True/False
+            match: a function (usually `any` or `all`) that accepts an iterable
+                and returns True/False. Used to aggregate the outputs of each
+                filter to decide whether a match has occurred. The default is
+                `any`
+        '''
+        new_ele_set = self._filter_elements(filters, match)
+        keys = [key for key in self.__elements__ if key not in new_ele_set]
+        self.delete_inds(keys)
```
brimoor: `delete_keys`

benjaminpkane:
After pondering some issues, I'd like to remove the `backing_dir` as a concept to the user and always use a self-managed temp directory for both `BigContainer` and `BigSet`. `write_json` would write out both the index and copy the element jsons to the index's base directory, and `from_json` would expect such a structure. Archiving would not change.

This work isn't critical to the sprint, so its not a priority. Come Monday this PR will be ready.

brimoor:
Not necessarily opposed, but I'd be curious to hear your reasoning and what issues you encountered.

One halfway option is to give the user _optional_ control over the backing directory and default to the self-managed temp directory if they don't override it.

The copying of the backing directory to the index's base directory when serializing is the sticky part. I can imagine why you might be proposing that if you're trying to use `BigContainer` in an ETA pipeline. The pipeline system would need a way to automatically generate paths to work with BigContainers.

Perhaps `to_archive` should support an "uncompressed mode" where the data is just written to a directory without any archiving?

benjaminpkane:
Most 

> Not necessarily opposed, but I'd be curious to hear your reasoning and what issues you encountered.
> 
> One halfway option is to give the user _optional_ control over the backing directory and default to the self-managed temp directory if they don't override it.
> 
> The copying of the backing directory to the index's base directory when serializing is the sticky part. I can imagine why you might be proposing that if you're trying to use `BigContainer` in an ETA pipeline. The pipeline system would need a way to automatically generate paths to work with BigContainers.
> 
> Perhaps `to_archive()` should support an "uncompressed mode" where the data is just written to a directory without any archiving?

Having the option might be better. My reasoning is that if `backing_dir` is required methods on `ImageLabels` (and the `VideoLabels` analogs) like `get_matches`, `count_matches`, `extract_image_labels`, and `copy` need to be overridden for `BigImageSetLabels` to accept a `backing_dir`.

It's not that much work, but seems like a troublesome practice for implementations in general. `BigContainer`s and `BigSet`s are also less interchangeable with the notion of a `backing_dir` also being present for any copy-like operations.

It might also be nice if `ImageSetLabels` was a direct subclass of `Set` (as opposed to the `images` attribute).

brimoor:
I definitely agree that life would be much better if `ImageSetLabels` and `VideoSetLabels` were direct subclasses of `Set`! I didn't do it when I added `Set` because it will involve renaming some methods that are probably used elsewhere in our code.

But it would be waaaay better if this change were made, and `ImageSetLabels` and `VideoSetLabels` are relatively new, so I think now is the time to make this change :)

benjaminpkane:
Sounds good! I will start with making `ImageSetLabels` and `VideoSetLabels` direct subclasses of `Set`.

benjaminpkane:
The slicing PR has been merged into this branch.

`*SetLabels` are now direct subclasses of `Set`/`BigSet`.

The temp storage optional feature has only been added to `BigSet`. 

I've tested with `ImageSetLabels` and `BigImageSetLabels`, but no with their video counterparts. A few more things to do like `to_set` for `BigSet`.

PR #281, opened 2019-07-31
process method for Models

benjaminpkane:
DRAFT. Not ready for review.

Adds a concrete method `process` method to all abstract model types with an associated input data type (`video_path` i.e. `VideoModel`, `imgs` tensor/list i.e. `VideoFramesModel`, `img` i.e. `ImageModel`).

####  A detail of note:
A `VideoFrameModel`'s `process` method returns a `VideoLabels` with frames indexed by the order of the tensor/list. So, in most cases, the `VideoLabels` return by `VideoFrameLabels.process(imgs)` are more of an intermediate instance. Best I could come up with.

### DIFF
--- a/eta/core/learning.py
+++ b/eta/core/learning.py
@@ -302,7 +302,7 @@ class VideoFramesModel(Model):
             imgs: a list (or d x ny x nx x 3 tensor) of images to process
 
         Returns:
-            an `eta.core.image.VideoLabels` instance containing the labels
+            an `eta.core.video.VideoLabels` instance containing the labels
                 generated for the given tensor of images
         '''
         raise NotImplementedError("subclasses must implement process()")
@@ -502,7 +502,7 @@ class VideoFramesClassifier(Classifier, VideoFramesModel):
             imgs: a list (or d x ny x nx x 3 tensor) of images to process
 
         Returns:
-            an `eta.core.image.VideoLabels` instance containing the labels
+            an `eta.core.video.VideoLabels` instance containing the labels
                 generated for the given tensor of images
         '''
         labels = etav.VideoLabels()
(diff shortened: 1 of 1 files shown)


eta/core/learning.py
```diff
@@ -443,6 +495,24 @@ def predict(self, imgs):
...
+        Args:
+            imgs: a list (or d x ny x nx x 3 tensor) of images to process
+
+        Returns:
+            an `eta.core.video.VideoLabels` instance containing the labels
+                generated for the given tensor of images
+        '''
+        labels = etav.VideoLabels()
+        attrs = self.predict(imgs)
+        for idx, imgs in enumerate(imgs):
```
brimoor: `enumerate(imgs, 1)` uses 1-based indexing

eta/core/learning.py
```diff
@@ -623,6 +717,24 @@ def detect(self, imgs):
...
+
+        Args:
+            imgs: a list (or d x ny x nx x 3 tensor) of images to process
+
+        Returns:
+            an `eta.core.video.VideoLabels` instance containing the labels
+                generated for the given tensor of images
+        '''
+        labels = etav.VideoLabels()
+        attrs = self.detect(imgs)
```
brimoor: This returns a `DetectedObjectContainer`, not an `AttributeContainer`.

As you pointed out, its not quite clear how to embed these results in a `VideoLabels` because "video frames" models output intermediate results that are used in the context of a larger algorithm. We should chat so I can understand more about what your use case is for this.

For example, the idea is that a `VideoFramesObjectDetector` would be responsible for including `DetectedObjects` with frame numbers to indicate which frame of the input tensor the detection corresponds to. So, this method should partition the `DetectedObjects` into their respective frames.

With `VideoFramesClassifier`, your strategy of inserting the attributes into each frame _is_ the intended interpretation of the output of that model.

benjaminpkane: Late response, apologies. I can chat whenever. And thanks for you thoughts!

I think I may just close this PR. The intention was to have a common interface for Models, but it is not turning out to be that useful.

PR #286, opened 2019-08-09
Sample Videos to Images Update

mattphotonman:
Changes sample_videos_to_images() so that you can specify a `stride` and `num_images`.  This will sample frames at the given `stride` until you reach the requested number of images.  One side effect of the implementation is that even when you don't specify `stride`, the function will never output more than `num_images` as it may have before this change.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -255,8 +255,10 @@ def sample_videos_to_images(
         stride: optional frequency with which to sample frames from
             videos
         num_images: optional total number of frames to sample from
-            videos. Only one of `stride` and `num_images` can be
-            specified at the same time.
+            videos. If `stride` is not specified then it will be
+            calculated based on `num_images`. If `stride` is
+            specified, frames will be sampled at this stride until
+            a total of `num_images` are obtained.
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -275,7 +277,7 @@ def sample_videos_to_images(
 
     _validate_stride_and_num_images(stride, num_images)
 
-    if num_images is not None:
+    if num_images is not None and stride is None:
         stride = _compute_stride(video_dataset, num_images, frame_filter)
         logger.info("Sampling video frames with stride %d", stride)
```
rpinnaka: Can the logger statement be changed? This is called when stride is None and there is no value for stride in `logger.info("Sampling video frames with stride %d", stride)`

mattphotonman: Not sure what you mean @rpinnaka . The stride is set in the line above this.

rpinnaka: @mattphotonman If the stride is originally set by the user, then the logger message will not be showing up. Can this be included for this case as well (which will be helpful for the user)?

PR #292, opened 2019-09-05
Dataset transformers update

mattphotonman:
Made some updates to dataset transformers, especially `Balancer`, and added some new dataset transformers.


eta/core/datasets.py
```diff
@@ -1482,8 +1492,8 @@ def __init__(self, data_path, labels_path):
             data_path (str): path to data file
             labels_path (str): path to labels json
         '''
-        self._data_path = data_path
-        self._labels_path = labels_path
+        self.data_path = data_path
```
tylerganter: The `__init__` docstring above reads "The label and data paths cannot and should not be modified after initialization." I think you should be creating and removing records, rather than modifying in place.

eta/core/datasets.py
```diff
@@ -1797,6 +1820,14 @@ def transform(self, src):
         for ki in keep_idxs:
             src.add(old_records[record_idxs[ki]])
 
+    def _validate(self):
+        if all([self.attr_name is None,
```
tylerganter: either `attr_name` or `labels_schema` must be provided. `object_label` is used with `attr_name` if provided, but is not a sufficient input on its own.

eta/core/datasets.py
mattphotonman: I don't modify this in place.  The reason for making these variables public is so that they get serialized when the `BuilderDataset` is written to JSON.

eta/core/datasets.py
```diff
@@ -1943,11 +2000,12 @@ def _to_helper_list_video_objects(self, records):
             NO_ID = 'NO_ID'
             helper_dict = defaultdict(set)
 
-            for frame_no, frame in labels.frames.items():
+            for frame_no in labels:
                 if (frame_no < record.clip_start_frame
                         or frame_no >= record.clip_end_frame):
                     continue
 
+                frame = labels[frame_no]
```
tylerganter: is this an optimization? why was this changed?

eta/core/datasets.py
mattphotonman: Yes, that is verified [here](https://github.com/voxel51/eta/blob/7f1cd8c483bff7adb20ce56e8472c15486310773/eta/core/datasets.py#L1909-L1912).  Are you suggesting we move that check into the `_validate()` function?

eta/core/datasets.py
mattphotonman: Syntax cleanup.

eta/core/datasets.py
tylerganter: Well, I think you could get rid of that second check, and do this:

```
if ((self.attr_name is not None and self.labels_schema is not None)
                or (self.attr_name is None and self.labels_schema is None)):
            raise ValueError(
                "One and only one of the following must be specified: "
                "attr_name, labels_schema")

        if self.object_label is not None and self.attr_name is None:
            raise ValueError(
                "attribute_name must be specified if object_label is "
                "specified")
```

mattphotonman: Ok, I changed things around a bit, is this clearer now?

eta/core/datasets.py
mattphotonman: Fixed.

eta/core/datasets.py
tylerganter: yep!

PR #296, opened 2019-09-09
Function to find duplicate files in a directory

mattphotonman:
- Added function `eta.core.utils.find_duplicate_files()`
- Used this to add a `deduplicate()` method for the `LabeledDataset` class

Note that because `deduplicate()` is a method of `LabeledDataset`, I wrote it so that it only changes the internal index.  If you want to actually delete duplicate files and update the manifest, you would have to do something like

```python
labeled_dataset.deduplicate()
labeled_dataset.prune()
labeled_dataset.write_manifest("manifest.json")
```

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -781,6 +781,64 @@ class LabeledDataset(object):
         merged_dataset.write_manifest(merged_dataset_path)
         return merged_dataset
 
+    def deduplicate(self):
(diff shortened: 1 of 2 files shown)


mattphotonman:
> Very cool implementation!

Thanks!  Algorithm fun!  But sorry to do this, I just made some changes and added a method for de-duplicating `LabeledDataset`s, so please re-review.

PR #297, opened 2019-09-11
Using frombuffer when reading video frames, for efficiency

brimoor:
`fromstring` works, but it copies the data. `frombuffer` avoids copying the data:
https://stackoverflow.com/questions/22236749/numpy-what-is-the-difference-between-frombuffer-and-fromstring

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -2368,7 +2368,7 @@ class FFmpegVideoReader(VideoReader):
 
         width, height = self.frame_size
         try:
-            vec = np.fromstring(self._raw_frame, dtype="uint8")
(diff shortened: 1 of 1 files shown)


eta/core/video.py
```diff
@@ -2368,7 +2368,7 @@ def _retrieve(self):
 
         width, height = self.frame_size
         try:
-            vec = np.fromstring(self._raw_frame, dtype="uint8")
+            vec = np.frombuffer(self._raw_frame, dtype="uint8")
```
mattphotonman: Noting that the array output by `_retrieve()` will now be read-only with this change.

PR #298, opened 2019-09-11
Symlink data directory in dataset

mattphotonman:
Adding the option to symlink the `data/` directory instead of copying over all of the files, in the `LabeledDataset.copy()` method.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -723,7 +723,8 @@ class LabeledDataset(object):
             os.rmdir(new_data_subdir)
             os.rmdir(new_labels_subdir)
             os.symlink(
-                os.path.join(self.data_dir, self._DATA_SUBDIR),
+                os.path.abspath(os.path.realpath(
+                    os.path.join(self.data_dir, self._DATA_SUBDIR))),
                 new_data_subdir
             )
             shutil.copytree(
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -31,6 +31,7 @@ import logging
 import os
 import random
 import re
+import shutil
 
 import numpy as np
 
@@ -691,7 +692,7 @@ class LabeledDataset(object):
 
         return self
 
-    def copy(self, dataset_path):
+    def copy(self, dataset_path, symlink_data=False):
         '''Copies the dataset to another directory.
 
         If the dataset index has been manipulated, this will be reflected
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -712,10 +719,24 @@ def copy(self, dataset_path):
...
-            etau.copy_file(labels_path, new_labels_subdir)
+        if symlink_data:
+            os.rmdir(new_data_subdir)
+            os.rmdir(new_labels_subdir)
+            os.symlink(
+                os.path.abspath(os.path.realpath(
+                    os.path.join(self.data_dir, self._DATA_SUBDIR))),
+                new_data_subdir
+            )
+            shutil.copytree(
```
tylerganter: Why do this instead of copying the labels the same as before? And do we want to symlink the directory, or the individual data files? I would think that using file symlinks would help keep track of any modification to the original directory

mattphotonman: Wanted to add the option to symlink in general because often I want to modify just the labels, but I want to save the original labels, so I don't want to copy over the entire data directory in that case if it's large.
Symlinking the individual files could work, but I was also worried that would take prohibitively long if there are a lot of files.  No perfect solution here.  Maybe we could add the option to either symlink the individual files or the data directory?

tylerganter: No I don't think it needs to be an option, let's just make a decision. But why would it take long to create symlinks for each video? Each of those many files being created is very small.

The second question was related to labels though, why copy using `copytree` instead of the same method used below?

mattphotonman: Let me test it I guess? The current training dataset has ~150K images, and if we took more frames from videos we could have as many as ~1.5 million.  I'm wondering if 1.5 million commands to make a symlink will start to add up.
The motivation for using `copytree()` is similar to the above: I'm definitely finding that when I make a copy of datasets, or when I merge them, it takes long time using these built-in methods, compared to if I just copy whole directories using shell commands.  My hypothesis is that this is due to overhead for each individual file copy command.

mattphotonman: Symlinking directory vs. files: It doesn't take longer to create symlinks to the invidual files, _however_, there are issues once a directory containing a large number of symlinks is created.  For example, on bb1 go to `/data1/temp/matthew/symlinked_data_files/data` and type `ls` and see how long it takes.  Compare this to doing the same in the original directory `/scratch/data/roadsense_training_datasets/train_image_dataset/data`. Also, note the size difference between a directory full of symlinks and a directory that is just a symlink to another directory:
```bash
$ du -sh /data1/temp/matthew/symlinked_data_*/data
4.0K    /data1/temp/matthew/symlinked_data_dir/data
586M    /data1/temp/matthew/symlinked_data_files/data
```

mattphotonman: **Copying the full `labels/` directory vs. copying individual files**: It doesn't take longer to copy individual labels files than it does to use `copytree()` to copy the entire directory.  However, if we choose to symlink the full `data/` directory, I would argue that we should copy the full `labels/` directory too for consistency.  (If you copy file-by-file, iterating over `self.iter_labels_paths()`, you will only copy the labels files in `self.dataset_index`, not necessarily all of the labels in the directory.)

tylerganter: Okay I'm convinced. I'm definitely baffled at how large the directory of symlinks is, even with the size of the dataset. Big data indeed

mattphotonman: Yeah, I don't quite get it either.  I know that unix does have trouble with large numbers of file handles in general.

PR #300, opened 2019-09-13
Added new balancer algorithm 'simple'

mattphotonman:
This algorithm crushes it for the one example I ran!

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -1816,6 +1816,7 @@ class Balancer(DatasetTransformer):
         `object_label`.
     '''
 
+    _NUM_RANDOM_ITER = 10000
     _BUILDER_RECORD_TO_SCHEMA = [
         (BuilderImageRecord, etai.ImageLabelsSchema),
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -2358,6 +2361,69 @@ def _greedy(self, A, b):
...
+            # descending order and drop elements as long as that won't
+            # make the sum go below attr_target.
+            new_dropped_counts = dropped_counts[attr_idx]
+            indices_to_remove = []
+            for record_idx in np.flip(np.argsort(dropped_counts_for_attr)):
+                if new_dropped_counts <= attr_target:
+                    break
+
+                count = dropped_counts_for_attr[record_idx]
+                if count ==0 or new_dropped_counts - count < attr_target:
```
tylerganter: add a space after `==`

mattphotonman:
> Awesome work! Simple indeed. Just a typo, but otherwise looks good

It's interesting to consider what we actually want in a balancing algorithm.  For example, this new one may yield a much smaller number of files, meaning less diversity.

PR #308, opened 2019-09-25
Adding method to split video into (roughly) equal-sized chunks

brimoor:


eta/core/video.py
```diff
@@ -1980,6 +1980,62 @@ def extract_keyframes(video_path, output_patt=None):
...
+    if num_clips:
+        # Round up to nearest second to ensure that an additional small clip
+        # is not created at the end
+        segment_time = np.ceil(metadata.duration / num_clips)
+    elif clip_duration:
+        segment_time = clip_duration
+    elif clip_size_bytes:
+        # Round up to nearest second to ensure that an additional small clip
+        # is not created at the end
+        segment_time = np.ceil(metadata.size_bytes / clip_size_bytes)
```
mattphotonman: Wouldn't this give the number of clips rather than the segment time in seconds?

brimoor: oops good point. fixed!

PR #316, opened 2019-10-08
Update virtualenv_guide.md

iantimmis:
We require Python 3.6.X, as Python 3.7+ causes issues. Because of this, I have updated the virtualenv_guide.md document with instructions on how to install Python 3.6.X on a fresh machine, or revert back to a previous version of Python if you have had it installed before.

### DIFF
--- a/docs/virtualenv_guide.md
+++ b/docs/virtualenv_guide.md
@@ -24,13 +24,15 @@ brew install python
 brew install python3
 ```
 * <b>Note:</b>
-Since we currently only support up to Python 3.6.X, you may need to brew install this older Python version. If you have had Python 3.6 installed previously, run the following:
+Since we currently only support Python 3.6.X, you may need to install this older Python version. Homebrew allows us to roll back to a previous version of Python. If, for example, you have had Python 3.6.5 installed previously, you may run the following:
 ```shell
+# Mac
 brew switch python 3.6.5_1
 ```
 
 * If you need a fresh install of Python 3.6.5, run the following:
 ```shell
+# Mac
 brew unlink python
 brew install pkg-config gdbm openssl readline sqlite xz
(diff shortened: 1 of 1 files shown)


docs/virtualenv_guide.md
```diff
@@ -23,6 +23,18 @@ sudo apt-get install python3.6
 brew install python
 brew install python3
 ```
+* <b>Note:</b>
+Since we currently only support up to Python 3.6.X, you may need to brew install this older Python version. If you have had Python 3.6 installed previously, run the following:
```
mattphotonman: I would say that we support Python 3.6.X, not "up to", since we don't support python 3.5.X for example.

mattphotonman: It looks like this command is specifically for python 3.6.5, so I would say, "if you have python 3.6.5 installed previously".

docs/virtualenv_guide.md
```diff
@@ -23,6 +23,18 @@ sudo apt-get install python3.6
...
+```shell
+brew switch python 3.6.5_1
+```
+
+* If you need a fresh install of Python 3.6.5, run the following:
+```shell
+brew unlink python
+brew install pkg-config gdbm openssl readline sqlite xz
+brew install --ignore-dependencies https://raw.githubusercontent.com/Homebrew/homebrew-core/f2a764ef944b1080be64bd88dca9a1d80130c558/Formula/python.rb
+```
```
mattphotonman: Please note somewhere that these instructions apply to Mac OS specifically.

iantimmis: Good feedback. I made a commit with updates you suggested. Also may be good to remember that we don't have a documented solution for linux users ATM

lethosor:
I've found [pyenv](https://github.com/pyenv/pyenv) to be useful for installing Python versions that aren't provided by Linux package managers. It works on macOS too, although it requires compiling Python from source, which I think Homebrew handles a little better (so it's probably worth keeping the Homebrew instructions).

PR #331, opened 2019-11-04
TempDir: creating base directory if necessary

brimoor:
This PR basically just adds `eta.core.utils.ensuredir()` to `eta.core.utils.make_temp_dir()` to ensure that the base directory exists before creating it.

This will make it so that repeated calls to `TempDir` won't cause any errors due to the base directory originally existing but now not 

Note that it _is_ intentional for `__exit__()` to use `eta.core.utils.delete_dir()`, which will prune the directory tree as possible to avoid empty directories sticking around

### DIFF
--- a/eta/core/utils.py
+++ b/eta/core/utils.py
@@ -1646,32 +1646,34 @@ class MD5FileHasher(FileHasher):
             return str(hashlib.md5(f.read()).hexdigest())
 
 
-def make_temp_dir(dir=None):
+def make_temp_dir(basedir=None):
     '''Makes a temporary directory.
 
     Args:
-        dir: an optional directory in which to put the temp dir
+        basedir: an optional directory in which to create the new directory
 
     Returns:
         the path to the temporary directory
     '''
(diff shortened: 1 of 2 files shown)


eta/core/utils.py
```diff
@@ -1646,32 +1646,34 @@ def hash(path):
...
     Args:
-        dir: an optional directory in which to put the temp dir
+        basedir: an optional directory in which to create the new directory
 
     Returns:
         the path to the temporary directory
     '''
-    return tempfile.mkdtemp(dir=dir)
+    if basedir:
+        ensure_dir(basedir)
```
tylerganter: This actually isn't going to fix the problem. If `/tmp` doesn't exist, then:
```
tempfile.mkdtemp()
```
will fail

brimoor: what? `make_temp_dir("/tmp")` will make a directory `"/tmp/xxx"`, so the first step is to ensure that `"/tmp"` exists

PR #334, opened 2019-11-08
Average Video Duration for `LabeledVideoDataset`

tylerganter:


eta/core/datasets.py
```diff
@@ -1372,6 +1372,18 @@ def validate_dataset(cls, dataset_path):
             if not os.path.isfile(labels_path):
                 raise LabeledDatasetError("File not found: %s" % labels_path)
 
+    @property
+    def average_video_duration(self):
```
brimoor: My two cents: expensive operations like this should be methods, not properties. Making something a property implies to the user that it is fast and safe to access many times. This could require reading 1000's of files from disk to compute.

I'd recommend `compute_average_video_duration()`

mattphotonman: Agreed.

tylerganter: Yeah I was thinking about this same thing when I wrote it. I'll keep this convention for properties going forward. Fixed

PR #336, opened 2019-11-11
Migrate to Tensorflow 2.0

tylerganter:
Making this a draft for now. Once this happens there's no going back!


iantimmis:
Love the initiative

tylerganter:
Closing because this doesn't need to be sitting around in the repo, and first we should probably focus on using `tf.compat`

tylerganter:
Closing because this doesn't need to be sitting around in the repo, and first we should probably focus on using `tf.compat`

PR #337, opened 2019-11-11
Handle `ffmpeg` errors in `sample_select_frames()`

mattphotonman:


eta/core/video.py
```diff
@@ -1908,7 +1888,22 @@ def sample_select_frames(
...
-        ffmpeg.run(video_path, tmp_patt)
+
+        #
+        # Make sure to handle case where ffmpeg command errors out.
+        # This can happen if the `frames` argument passed to the current
+        # function includes frames that are not actually present in the video.
+        #
+        try:
+            ffmpeg.run(video_path, tmp_patt)
+        except etau.ExecutableRuntimeError as e:
```
brimoor: I'm not comfortable approving this as a fix until we understand what exactly was causing problems in the first place. How do we know that switching to `VideoProcessor` will change anything? If the problem is that `ffmpeg` can't parse some frame of the video, that will be true with both approaches.

Secondly, if failures in this ffmpeg command will result in a partially populated temp dir, then that's all we want anyway. There is no need to redo all the expensive computation with `VideoProcessor` just to achieve the same result.

As you can tell, I don't like approving bandaid PRs in our core methods in ETA. We need bulletproof code here!

eta/core/video.py
```diff
@@ -1908,7 +1888,22 @@ def sample_select_frames(
...
+            err_msg = str(e)
+            if len(err_msg) > 350:
+                err_msg = "%s ... %s" % (err_msg[:175], err_msg[-175:])
+            logger.info("FFmpeg runtime error: '%s'", err_msg)
```
brimoor: you should use `logger.error` to log errors

eta/core/video.py
```diff
@@ -1908,7 +1888,22 @@ def sample_select_frames(
...
+        #
+        # Make sure to handle case where ffmpeg command errors out.
+        # This can happen if the `frames` argument passed to the current
+        # function includes frames that are not actually present in the video.
+        #
+        try:
+            ffmpeg.run(video_path, tmp_patt)
+        except etau.ExecutableRuntimeError as e:
+            err_msg = str(e)
+            if len(err_msg) > 350:
```
brimoor: why? just print the whole error?

eta/core/video.py
```diff
@@ -1854,46 +1854,26 @@ def sample_select_frames(
     Returns:
         a list of the sampled frames if output_patt is None, and None otherwise
     '''
-    # Parse parameters
-    resize_images = size is not None
+    if fast:
```
brimoor: If we are going to have logic that handles failures, I believe it should be here, not inside the private methods.

It's confusing that, as written, `_sample_select_frames_fast()` is really "try to do it fast, but there are two different cases in which fast won't work and we will secretly call `_sample_select_frames_slow()` here

eta/core/video.py
```diff
@@ -1908,7 +1891,20 @@ def sample_select_frames(
...
+
+        try:
+            ffmpeg.run(video_path, tmp_patt)
+        except etau.ExecutableRuntimeError:
+            #
+            # Sometimes ffmpeg can't decode frames in video. Analogous to
+            # FFmpegVideoReader, our approach here is to gracefully fail and
+            # just give the user however many frames we can...
+            #
+            num_frames = len(etau.parse_pattern(tmp_patt))
```
tylerganter: awesome!

mattphotonman:
> LGTM. Note that I removed the min fraction of frames business as a mechanism to fallback to "slow" mode. The reason is that both commands are using `ffmpeg` and there is no fundamental reason to expect that "slow" mode as any greater change of success than "fast" mode.
> 
> I added a comment explaining that we are taking the same approach as `FFmpegVideoReader` by gracefully failing and returning whatever frames we can to the user.

Great, thanks for doing that!

PR #339, opened 2019-11-12
Maximum video size (bytes or seconds)

tylerganter:
Adding a optional parameters that check video size and duration and throw an error without attempting to process if these checks fail.

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1608,7 +1608,7 @@ class VideoStreamInfo(Serializable):
         except KeyError:
             pass
 
-        logger.warning("Unable to determine filesize; returning -1")
+        logger.warning("Unable to determine video size; returning -1")
         return -1
 
     def attributes(self):
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1475,16 +1475,15 @@ class BigVideoSetLabels(VideoSetLabels, BigSet):
 class VideoStreamInfo(Serializable):
     '''Class encapsulating the stream info for a video.'''
 
-    def __init__(self, stream_info, format_info=None):
+    def __init__(self, stream_info, format_info):
         '''Constructs a VideoStreamInfo instance.
 
         Args:
             stream_info: a dictionary of video stream info
-            format_info: an optional dictionary of video format info. By
-                default, no format info is stored
+            format_info: a dictionary of video format info
         '''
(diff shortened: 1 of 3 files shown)


eta/modules/sample_videos.py
```diff
@@ -37,6 +37,7 @@
 import eta.core.image as etai
 import eta.core.module as etam
 import eta.core.video as etav
+import eta.core.utils as etau
```
brimoor: nit: style says alphabetize imports. 

ETA has a style guide and we follow it https://github.com/voxel51/eta/blob/develop/docs/python_style_guide.md.

eta/modules/sample_videos.py
```diff
@@ -182,6 +197,27 @@ def _process_video(input_path, output_frames_dir, parameters):
         input_path, frames, output_patt=output_patt, size=size, fast=True)
 
 
+def _check_input_video_size(
+        stream_info, max_video_file_size, max_video_duration):
+    if max_video_file_size is not None:
+        file_size = int(stream_info.format_info["size"])
```
brimoor: you should use `stream_info.size_bytes` for this. Use properties whenever possible

eta/modules/sample_videos.py
```diff
@@ -182,6 +197,27 @@ def _process_video(input_path, output_frames_dir, parameters):
...
+        if file_size > max_video_file_size:
+            raise ValueError(
+                "Input video file size must be less than {}; found {}".format(
+                    etau.to_human_bytes_str(max_video_file_size),
+                    etau.to_human_bytes_str(file_size))
+            )
+
+    if max_video_duration is not None:
+        video_dur = float(stream_info.format_info["duration"])
```
brimoor: you should use `stream_info.duration` to get this. Use properties whenever possible

eta/core/video.py
```diff
@@ -1599,19 +1598,18 @@ def duration(self):
...
-        Args:
-            key: the key to lookup in the stream info dictionary
-
-        Returns:
-            the value for the given key
-
-        Raises:
-            KeyError: if the key was not found in the stream info dictionary
+    @property
+    def size_bytes(self):
```
brimoor: @tylerganter `size_bytes` was not exposed as a property of `VideoStreamInfo`, but it should have been, so I added it here

PR #343, opened 2019-11-20
Consistent filenames after dataset transforms

tylerganter:
Dataset builder now preserves the data and labels filenames.

If filenames conflict an incrementing integer is appended until a unique name is found. Example output from my test with 3 records with filename `cat.10649.jpg`:
```
cat.10649-0.jpg
cat.10649-1.jpg
cat.10649.jpg
```
This was branched off of `wrangle-helpers` but so I'm merging it into there. There are two PRs so that it's easier to review each individually

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -3011,15 +3011,3 @@ class FilterByPath(DatasetTransformer):
 class DatasetTransformerError(Exception):
     '''Exception raised when there is an error in a DatasetTransformer'''
     pass
-
-
-class TylersTestTransform(DatasetTransformer):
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -661,6 +661,20 @@ def split(self, split_fractions=None, descriptions=None,
 
         return dataset_list
 
+    def has_data_with_name(self, data_path):
+        '''Checks whether if a data file already exists in the dataset with the
```
iantimmis: This is awkward. Could you remove the word "if"

PR #346, opened 2019-11-20
New `remove_objects_without_attrs` function

tylerganter:
Breaking this out from `filter_by_schema` and adding some modularity by specifying which labels to remove.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -2777,19 +2777,20 @@ class SchemaFilter(DatasetTransformer):
     is None, no filtering is done.
     '''
 
-    def __init__(self, schema, remove_objects_without_attrs=False,
-                 prune_empty=True):
+    def __init__(self, schema, object_labels_to_filter=None, prune_empty=True):
         '''Initialize the SchemaFilter with a schema.
 
         Args:
             schema (VideoLabelsSchema or ImageLabelsSchema)
-            remove_objects_without_attrs: (bool) If True, remove objects that no
-                longer have attributes after filtering
+            object_labels_to_filter: a list of DetectedObject.label strings that
+                each object instance with no attributes is removed.
+                DetectedObject.labels not in this list are not removed even if
+                they have no attributes.
             prune_empty (bool) if True, records whose labels are empty after
                 filtering are pruned from the dataset.
         '''
         self.schema = schema
-        self.remove_objects_without_attrs = remove_objects_without_attrs
+        self.object_labels_to_filter = object_labels_to_filter
         self.prune_empty = prune_empty
 
     def transform(self, src):
@@ -2808,8 +2809,10 @@ class SchemaFilter(DatasetTransformer):
         src.clear()
         for record in old_records:
             labels = record.get_labels()
-            labels.filter_by_schema(self.schema,
-                                    self.remove_objects_without_attrs)
+            labels.filter_by_schema(self.schema)
+            if self.object_labels_to_filter is not None:
+                labels.remove_objects_without_attrs(
+                    self.object_labels_to_filter)
             if not self.prune_empty or not labels.is_empty:
                 record.set_labels(labels)
                 src.add(record)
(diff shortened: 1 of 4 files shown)


eta/core/datasets.py
```diff
@@ -2751,20 +2751,21 @@ class SchemaFilter(DatasetTransformer):
     is None, no filtering is done.
     '''
 
-    def __init__(
-            self, schema, remove_objects_without_attrs=False,
-            prune_empty=True):
-        '''Initialize the SchemaFilter with a schema.
+    def __init__(self, schema, object_labels_to_filter=None, prune_empty=True):
```
brimoor: With this refactor you'd have to set `object_labels_to_filter = []` if you wanted to NOT filter objects with no attributes. That's a bit confusing. I think you still need the `remove_objects_without_attrs` flag here to more intuitively control whether you want filtering to happen

tylerganter: True, I added another boolean param to help clarify thtis

eta/core/image.py
```diff
@@ -174,17 +174,25 @@ def merge_labels(self, image_labels):
...
+
+    def remove_objects_without_attrs(self, labels=None):
+        '''Removes DetectedObjects from this instance that do not have
+        attributes.
+
+        Args:
+            labels: an optional list of object labels to which to restrict
+                attention when filtering. By default, all objects are processed
+        '''
+        self.objects.remove_objects_without_attrs(labels=labels)
```
brimoor: Note that I used the `labels=labels` syntax to invoke the `remove_objects_without_attrs` method.

When invoking methods with keyword arguments, please always use the `name=value` syntax. That way, if someone else comes along and adds a new keyword argument to the method that changes the order of the arguments, any existing invocations of that method will still work.

Technically, one should NOT insert keyword arguments anywhere besides the end of the list to guarantee backwards compatibility, but, hey, it happens. For example, you yourself added the new `object_labels_to_filter` argument to `SchemaFilter` BEFORE the existing `prune_empty` argument, which could break anyone's existing code that reads `SchemaFilter(schema, prune_empty)`

eta/core/datasets.py
```diff
@@ -2751,20 +2751,25 @@ class SchemaFilter(DatasetTransformer):
...
         Args:
             schema: a VideoLabelsSchema or ImageLabelsSchema
             remove_objects_without_attrs: whether to remove objects with no
-                attributes, after filtering. By default, this is False
+                attributes, after filtering. Use the `object_labels_to_filter`
+                argument to control which object labels are filtered. By
+                default, this is False
+            object_labels_to_filter: an optional list of object labels for
+                which to restrict attention when filtering objects with no
+                attributes. If None, all objects are filtered
```
brimoor: One other thing related to style- I see that you sometimes introduce docstrings that use the `parameter (type): description` syntax to document the type of the argument. I wish that all of ETA did that, as that would allow us to use Sphinx (as we do in https://github.com/voxel51/api-py) to generate auto-documentation that can parse out parameter types. However, we haven't adopted that in the existing codebase, so the style guide says to follow the existing docstring convention in the repo.

tl;dr: no `(type)` for now

eta/core/image.py
```diff
@@ -184,16 +184,15 @@ def filter_by_schema(self, schema):
...
+        '''Removes DetectedObjects from this instance that do not have
+        attributes.
 
         Args:
-            object_labels_to_filter: a list of DetectedObject.label strings.
-                DetectedObject.labels not in this list are not removed even if
-                they have no attributes. If None, all object labels are
-                filtered.
+            labels: an optional list of object labels to which to restrict
+                attention when filtering. By default, all objects are processed
```
tylerganter: I'm not keen on this description of the arg. I want to make it clear what each element of the list is, which is two things:
1) a string
2) a value that will be matched/compared against a `DetectedObject.label`
I'm indifferent to the exact syntax by which we convey that, but I think it makes this documentation a lot easier if these two snippets of information are conveyed.

brimoor: do what you will. PR is approved for you

PR #351, opened 2019-11-27
Improving docs

brimoor:
This PR adds two types of documentation:
(a) adds a description of `default_deployment_config_dict` to `docs/models_dev_guide.md`
(b) adds instructions for exporting frozen TF graphs for models from the TF-Slim framework

### DIFF
--- a/docs/export_tf_graphs_guide.md
+++ b/docs/export_tf_graphs_guide.md
@@ -1,0 +1,74 @@
+# Exporting TensorFlow Graphs Guide
+
+This document describes how to export frozen TensorFlow graphs for inference.
+
+
+## Exporting pre-trained models from the TF-Slim Model Zoo
+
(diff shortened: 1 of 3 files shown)


docs/export_tf_graphs_guide.md
```diff
@@ -0,0 +1,74 @@
+# Exporting TensorFlow Graphs Guide
+
+This document describes how to export frozen TensorFlow graphs for inference.
+
+
+## Exporting pre-trained models from the TF-Slim Model Zoo
```
brimoor: As pointed out by @iantimmis, this can be done much more succinctly via our existing codebase: https://github.com/voxel51/theta/blob/develop/theta/classifiers/tfslim_classifiers.py#L1166-L1179

I will update

PR #353, opened 2019-12-05
Prepend name fix

iantimmis:


eta/core/datasets.py
```diff
@@ -759,6 +759,11 @@ def add_file(self, data_path, labels_path, new_data_filename=None,
         new_labels_path = os.path.join(labels_subdir, new_labels_filename)
         if labels_path != new_labels_path:
             labels_method(labels_path, new_labels_path)
+            # renames the 'filename' attribute within the json file to reflect
+            # the updated data_filename
+            data = load_json(new_labels_path)
+            data['filename'] = os.path.basename(new_data_path)
+            write_json(data, new_labels_path)
```
mattphotonman: Better would be to use `self._read_labels()` instead of `load_json()`, because the former would read an actual `ImageLabels` or `VideoLabels` object. So you could do
```python
labels = self._read_labels(new_labels_path)
labels.filename = os.path.basename(new_data_path)
labels.write_json(new_labels_path)
```

tylerganter: good catch. yeah it's likely a red flag to even be importing `load_json` and `write_json` from `eta.core.serial`

eta/core/datasets.py
```diff
@@ -1842,9 +1847,11 @@ def build(self, path, description=None, pretty_print=False,
                 data_method=data_method
             )
 
-            # The `file_method` is irrelevant in this situation as the files
-            # are placed directly into the dataset by `record.build()`.
-            dataset.add_file(data_path, labels_path)
+            # MOVE here is used to rename the file
```
tylerganter: MOVE here is used
Use the force you must

eta/core/datasets.py
```diff
@@ -1842,9 +1847,11 @@ def build(self, path, description=None, pretty_print=False,
...
+            dataset.add_file(data_path, labels_path,
+                             os.path.basename(record.new_data_path),
+                             os.path.basename(record.new_labels_path),
```
benjaminpkane: I would recommend passing the final paths directly to `record.build()`. Labels are being loaded, having their `filename` attr changed ,and then written twice in the current form.

PR #355, opened 2019-12-10
Adding support for instance segmentation methods to inference framework

brimoor:
Change log
- adding `eta.detectors.TFModelsSegmenter` interface for performing inference with instance segmentation methods trained using the TF models detection framework
- adding support for storing instance masks (as serialized boolean numpy arrays) in `DetectedObject` instances
- adding support for visualizing instance masks on images/videos
- adding pre-trained Mask R-CNN models from TF models detection zoo to the ETA models registry

Example usage:
```py

from eta.detectors import TFModelsSegmenter, TFModelsSegmenterConfig
import eta.core.image as etai


TEST_IMAGE_PATH = "/path/to/image.jpg"
ANNO_IMAGE_PATH = "/path/for/annotated-image.jpg"
MODEL_NAME = "mask-rcnn-resnet101-atrous-coco"
CONFIDENCE_THRESH = 0.3

# Load model
config = TFModelsSegmenterConfig({
    "model_name": MODEL_NAME,
    "confidence_thresh": CONFIDENCE_THRESH,
})
segmenter = TFModelsSegmenter(config)

# Perform detection
img = etai.read(TEST_IMAGE_PATH)
with segmenter:
    objects = segmenter.detect(img)

# Visualize detections
image_labels = etai.ImageLabels(objects=objects)
etai.write(img_anno, ANNO_IMAGE_PATH)
```

Example output:
mask-rcnn-output


eta/detectors/tfmodels_segmenters.py
```diff
@@ -0,0 +1,197 @@
...
+
+    Note that this class implements the `HasDefaultDeploymentConfig` mixin, so
+    if a published model is provided via the `model_name` attribute, then any
+    omitted fields present in the default deployment config for the published
+    model will be automatically populated.
+
+    Attributes:
+        model_name: the name of the published model to load. If this value is
+            provided, `model_path` does not need to be
+        model_path: the path to a TF SavedModel in `.pb` format to load. If
```
iantimmis: To clarify, this would be `my_downloaded_model/saved_model/saved_model.pb`, and not the `my_downloaded_model/frozen_inference_graph.pb` ?

eta/detectors/tfmodels_segmenters.py
```diff
@@ -0,0 +1,197 @@
...
+            model_path = self.config.model_path
+
+        # Load model
+        self._tf_graph = self._build_graph(model_path)
+        self._sess = self.make_tf_session(graph=self._tf_graph)
+
+        # Load labels
+        label_map = gool.load_labelmap(self.config.labels_path)
+        categories = gool.convert_label_map_to_categories(
+            label_map, max_num_classes=90, use_display_name=True)
```
iantimmis: Where does this `90` come from?

eta/core/serial.py
```diff
@@ -181,6 +190,43 @@ def write_pickle(obj, path):
...
+def serialize_numpy_array(array):
+    '''Serializes a numpy array.
+
+    Args:
+        array: a numpy array
+
+    Returns:
+        the serialized string
+    '''
+    #return pickle.dumps(array)
```
iantimmis: There are 4 comments like this between `serialize_numpy_array` and `deserialize_numpy_array`. Can we remove these?

tylerganter:
Tested and works, great job! I added an example to the examples folder.

Reviewing the code now.

eta/core/objects.py
```diff
@@ -42,14 +43,15 @@ class DetectedObject(Serializable, HasBoundingBox):
...
-            frame_number=None, index_in_frame=None, attrs=None,
+            self, label, bounding_box, mask=None, confidence=None, index=None,
+            score=None, frame_number=None, index_in_frame=None, attrs=None,
             eval_type=None):
         '''Creates a DetectedObject instance.
 
         Args:
             label: object label string
             bounding_box: a BoundingBox around the object
+            mask: (optional) a mask for the object within its bounding box
```
tylerganter: What type is this?

eta/detectors/tfmodels_segmenters.py
brimoor: good catch. For reasons I don't understand, this method forces one to provide an upper bound on the number of classes to load from the label map, and 90 is the number from the COCO label map. I discovered that setting this to infinity works, so I did that.

90 = magic number. inf = cool.

eta/core/serial.py
brimoor: I converted these to a more proper comment. I was leaving bread crumbs for other serialization strategies that I considered

eta/core/objects.py
brimoor: clarified that it can be a numpy array of any kind (typically, 2D boolean)

eta/detectors/tfmodels_segmenters.py
brimoor: It is the frozen inference graph, not the saved model. Good catch. Updated.

PR #356, opened 2019-12-10
Labeled dataset fixes

mattphotonman:
I've refactored in a way such that everything should be consistent now, and tested.  Note that the unique appender is no longer needed in `LabeledDatasetBuilder.build()`, because this can be achieve by prepending the dataset name instead, so I took that section of the code out.  This also means that `dataset.add_file()` is now taking in the same path names as `record.build()`, so the `file_method` is again irrelevant.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -3013,9 +3013,9 @@ class PrependDatasetNameToRecords(DatasetTransformer):
         Args:
             src: a BuilderDataset
         '''
-        for i in range(len(src.records)):
-            base = _get_dataset_name(src.records[i].data_path)
-            src.records[i].prepend_to_name(prefix=base)
+        for record in src.records:
+            base = _get_dataset_name(record.data_path)
+            record.prepend_to_name(prefix=base)
 
 
 class FilterByFilename(DatasetTransformer):
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -710,7 +710,7 @@ class LabeledDataset(object):
 
     def add_file(self, data_path, labels_path, new_data_filename=None,
                  new_labels_filename=None, file_method=COPY,
-                 error_on_duplicates=False, update_filename_in_labels=True):
+                 error_on_duplicates=False):
         '''Adds a single data file and its labels file to this dataset.
 
         Args:
@@ -729,9 +729,6 @@ class LabeledDataset(object):
                 data file in the dataset. If this is set to `False`, the
                 previous mapping of the data filename to a labels file
                 will be deleted.
-            update_filename_in_labels: whether to update the `filename`
-                attribute inside the labels JSON, with the filename of the data
-                file being added by this method
 
         Returns:
             self
@@ -763,8 +760,8 @@ class LabeledDataset(object):
         if labels_path != new_labels_path:
             labels_method(labels_path, new_labels_path)
 
-        # Update the filename attribute in the labels JSON if requested
-        if update_filename_in_labels:
+        # Update the filename attribute in the labels JSON if necessary
+        if new_data_filename != os.path.basename(data_path):
             labels_ = self._read_labels(new_labels_path)
             labels_.filename = new_data_filename
             self._write_labels(labels_, new_labels_path)
@@ -784,7 +781,7 @@ class LabeledDataset(object):
         return self
 
     def add_data(self, data, labels, data_filename, labels_filename,
-                 error_on_duplicates=False, update_filename_in_labels=True):
+                 error_on_duplicates=False):
         '''Creates and adds a single data file and its labels file to this
         dataset, using the input python data structure.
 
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -1812,33 +1814,14 @@ def build(self, path, description=None, pretty_print=False,
...
-                    data_subdir,
-                    data_basename + unique_appender + data_ext
-                )
-                if dataset.has_data_with_name(data_path):
-                    continue
-                labels_path = os.path.join(
-                    labels_subdir,
-                    labels_basename + unique_appender + labels_ext
-                )
-                break
```
tylerganter: This is still relevant if:
- a video is split into clips
- two datasets with the same name are merged
- third thing that we haven't thought of

This isn't intended to be something is used, but it does act as a safeguard

eta/core/datasets.py
```diff
@@ -738,32 +738,33 @@ def add_file(self, data_path, labels_path, new_data_filename=None,
...
             labels_method(labels_path, new_labels_path)
-            # renames the 'filename' attribute within the json file to reflect
-            # the updated data_filename
-            data = load_json(new_labels_path)
-            data['filename'] = os.path.basename(new_data_path)
-            write_json(data, new_labels_path)
+
+        # Update the filename attribute in the labels JSON if necessary
+        if new_data_filename != os.path.basename(data_path):
+            labels_ = self._read_labels(new_labels_path)
```
tylerganter: `labels_`? what is this convention?

tylerganter:
Two small comments.

I couldn't really keep track of the changes and conflicts but I read through the `add_file` method and it looks good

eta/core/datasets.py
mattphotonman: The problem is that it always appends an index, even if you're not merging datasets at all, for example. This makes it a bit of pain to match filenames with the ones in the original dataset.  How about having logic that only adds indices once a duplicate name is found?

eta/core/datasets.py
mattphotonman: I wanted to use a variable name like `labels`, but was considered it was used elsewhere.  The underscore denotes a short-lived temp variable.

eta/core/datasets.py
mattphotonman: (If you agree with the approach of logic that only adds indices once a duplicate name is found, I can add that.)

eta/core/datasets.py
tylerganter: this is nit, I know, but isn't our convention a prepended underscore, not appended?

mattphotonman: I've seen both in our code base.  Prepend means it's a private variable.

eta/core/datasets.py
tylerganter: I don't know how or when this was changed. The original logic is here:
https://github.com/voxel51/eta/pull/343/files
It uses the method `has_data_with_name` in the while loop

tylerganter: It looks like the while loop needs to be fixed, and the `data_path` and `labels_path` should be initialized before entering the while loop, such that the while loop simply modifies `data_path` and `labels_path` if necessary.

benjaminpkane: These were my changes. Apologies.

eta/core/datasets.py
```diff
@@ -228,6 +228,46 @@ def _find_next_available_idx(idx, unavailable_indicators):
...
+    '''
+    if not dataset.has_data_with_name(data_path):
+        return data_path, labels_path
+
+    data_filename = os.path.basename(data_path)
+    labels_filename = os.path.basename(labels_path)
+    data_basename, data_ext = os.path.splitext(data_filename)
+    labels_basename, labels_ext = os.path.splitext(labels_filename)
+
+    filename_regex = re.compile("%s-([0-9]+)%s" % (data_basename, data_ext))
```
tylerganter: how fancyyyy

eta/core/datasets.py
```diff
@@ -1847,11 +1880,9 @@ def build(self, path, description=None, pretty_print=False,
...
             )
 
-            # MOVE here is used to rename the file
-            dataset.add_file(data_path, labels_path,
-                             os.path.basename(record.new_data_path),
-                             os.path.basename(record.new_labels_path),
-                             file_method=MOVE)
+            # The `file_method` is irrelevant because the files are already in
+            # the dataset directory
+            dataset.add_file(data_path, labels_path)
```
tylerganter: Is this because the copy is performed in the previous line? (`record.build`)

mattphotonman: Correct.  If the file already exists in the dataset directory, then `dataset.add_file()` doesn't do anything to files, it just adds them to the manifest.

eta/core/datasets.py
mattphotonman: I think we're all sorted out now!

PR #366, opened 2019-12-18
Faster RCNN NAS model added to registry

iantimmis:
Added Faster RCNN nas model to registry

### DIFF
--- a/eta/models/manifest.json
+++ b/eta/models/manifest.json
@@ -21,20 +21,40 @@
             "date_created": "2019-12-17 08:25:51"
         },
         {
-            "base_name": "faster_rcnn_nas_coco_2018_01_28",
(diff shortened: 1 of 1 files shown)


eta/models/manifest.json
```diff
@@ -20,6 +20,26 @@
             },
             "date_created": "2019-12-17 08:25:51"
         },
+        {
+            "base_name": "faster_rcnn_nas_coco_2018_01_28",
```
brimoor: Please follow the existing naming convention in this file. I.e., call this `faster-rcnn-nas-coco`.

Also, while you're here, please add faster-rcnn-nas-lowproposals-coco`. It's 3x faster, likely with similarly high accuracy.

PR #367, opened 2019-12-20
Adding support for extracting features in classification/detection modules

brimoor:
Release notes
- adds an `eta.core.features.FeaturesHandler` hierarchy of classes for reading/writing features
- adds types to `eta.core.types` to encapsulate the various types of features that can now be generated
- adds support for extracting features to all detection/classification modules
- adds an `eta.core.features.CachingVideoObjectsFeaturizer` to handle the featurization of detected objects in one or more videos
- `eta.classifiers.VGG16Classifier` now implements the `FeaturizingClassifier` mixin
- updated `eta.core.features.CachingVideoFeaturizer` to use `FeaturesHandler` behind-the-scenes


brimoor:
This PR is ready for review

eta/core/types.py
```diff
@@ -1391,7 +1596,7 @@ class ZippedVideoObjectsFeaturesDirectory(ZippedDirectory):
 
 
 class DataRecordsDirectory(JSONDirectory):
-    '''A directory containing a sequence of DataRecords JSON files.
+    '''A directoryof DataRecords JSON files.
```
benjaminpkane: Typo

brimoor: ty

brimoor:
@jasoncorso `eta.core.features` is ready for re-review.

All reading and writing of features is now handled via the `FeaturesHandler` classes. `BackingManger` is used purely for deciding on a backing directory to use.

I like this separation of tasks. I also like the existence of `FeaturesHandler` separate from the `Caching*` classes, because a common pattern is to write features to disk in one module, and then use them in one or more other modules, where it no longer matters what featurizer was used to generate the features, but rather only how they are stored on disk.

`Caching*` is updated to use these two constructs with no duplication of code.

Note that I switched from `.npz` to `.npy` because read/write is more of a bottleneck than disk space.

PR #372, opened 2019-12-21
Adding height/width/centroid to BoundingBox

brimoor:
Adding `height()`, `width()`, and `centroid()` methods to `eta.core.geometry.BoundingBox`.

Also adding return types to the existing docstrings in the module

### DIFF
--- a/eta/core/geometry.py
+++ b/eta/core/geometry.py
@@ -28,6 +28,11 @@ class BoundingBox(Serializable):
 
     The bounding box is represented as two RelativePoint instances describing
     the top-left and bottom-right corners of the box, respectively.
+
+    ETA follows the convention that the top-left corner of the image is [0, 0]
+    and the bottom-right corner of the image is [1, 1]. Thus, proper bounding
+    boxes satisfy the convention that their bottom-right coordinates are always
+    greater than their top-left coordinates.
     '''
 
     def __init__(self, top_left, bottom_right):
--- a/eta/core/geometry.py
+++ b/eta/core/geometry.py
@@ -31,7 +31,13 @@ class BoundingBox(Serializable):
     '''
 
     def __init__(self, top_left, bottom_right):
-        '''Constructs a BoundingBox from two RelativePoint instances.'''
+        '''Creates a BoundingBox instance.
+
+        Args:
+            top_left: a RelativePoint describing the top-left corner of the box
+            bottom_right: a RelativePoint describing the bottom-right corner of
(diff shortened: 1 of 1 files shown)


eta/core/geometry.py
```diff
@@ -129,11 +135,39 @@ def pad_relative(self, alpha):
...
-        '''Computes the area of the bounding box, in [0, 1].'''
-        w = self.bottom_right.x - self.top_left.x
-        h = self.bottom_right.y - self.top_left.y
-        return w * h
+        '''Computes the area of the bounding box, in [0, 1].
+
+        Returns:
+            the area
+        '''
+        return self.width() * self.height()\
```
tylerganter: Is this a typo? (the backslash)

eta/core/geometry.py
```diff
@@ -31,7 +31,13 @@ class BoundingBox(Serializable):
     '''
 
     def __init__(self, top_left, bottom_right):
-        '''Constructs a BoundingBox from two RelativePoint instances.'''
+        '''Creates a BoundingBox instance.
+
```
mattphotonman: It would be good if we could write down the conventions for "top" and "bottom", and for "right" and "left", with respect to `numpy` array indices.  This doc string would be a good place.  E.g. I find it hard to remember with the `x` coordinate for a "left" point should be greater or less than the `x` coordinate for a "right" point, and similarly for the `y` coordinate for a "top" vs. "bottom" point.

eta/core/geometry.py
```diff
@@ -129,11 +135,39 @@ def pad_relative(self, alpha):
 
         return BoundingBox(RelativePoint(tlx, tly), RelativePoint(brx, bry))
 
+    def height(self):
+        '''Computes the height of the bounding box, in [0, 1].
+
+        Returns:
+            the height
+        '''
+        return self.bottom_right.y - self.top_left.y
```
mattphotonman: For example, this ordering is unintuitive.

eta/core/geometry.py
brimoor: thanks, fixed

eta/core/geometry.py
brimoor: I added a discussion of this to the class docstring.

For the record, it would be far more confusing to adopt the opposite convention. We're simply following the same convention as NumPy, which loads images such that the top-left of the image is `[0, 0]` and the bottom-right is `[height, width]`

PR #374, opened 2019-12-23
Adding support for visualizing TF graph for a model

brimoor:
Now you can do this:

```
eta models --visualize-tf-graph faster-rcnn-inception-v2-coco

# open Tensorboard link printed by above command in browser to visualize the graph
```


brimoor:
@tylerganter comments are addressed. The model will be downloaded if necessary, and the flag was renamed to `--visualize-tf-graph` to hint to the user that this only works for TensorFlow graphs.

PR #378, opened 2019-12-27
Optionally set Alpha = Confidence in AnnotationConfig

tylerganter:
Fancy feature! Allows for the object confidence to be reflected in the transparency of the box

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -61,8 +61,8 @@ class AnnotationConfig(Config):
         linewidth: the linewidth, in pixels, of the object bounding boxes
         alpha: the transparency of the object bounding boxes and frame
             attributes
-        alpha_equals_confidence: True will set the transparency equal to the
-            object confidence
+        confidence_scaled_alpha: True will scale `alpha` and `mask_fill_alpha`
+            by the object confidence
         text_color: the annotation text color
         object_text_pad_pixels: the padding, in pixels, around the text in the
             object labels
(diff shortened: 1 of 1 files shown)


eta/core/annotations.py
```diff
@@ -430,12 +434,16 @@ def _annotate_object(img, obj, annotation_config):
...
     font = annotation_config.font
     alpha = annotation_config.alpha
+    alpha_equals_confidence = annotation_config.alpha_equals_confidence
     linewidth = annotation_config.linewidth
     pad = annotation_config.object_text_pad_pixels
     text_color = tuple(_parse_hex_color(annotation_config.text_color))
     mask_border_thickness = annotation_config.mask_border_thickness
     mask_fill_alpha = annotation_config.mask_fill_alpha
 
+    if alpha_equals_confidence and obj.confidence is not None:
```
brimoor: ooooh fancy! I haven't seen this visually yet, but I'm sure it will look cool, and also be an interesting way for less confident things to be less visually apparent. One potential issue is that different models' confidences mean different things. For some, 0.9999 means high confidence; for others, 0.4 means high confidence. So, we'll need to think about that in the future.

Right this minute, can I talk you into tweaking the implementation so that you multiply the existing alpha values by confidence rather than _replacing_ them with the confidence?

`AnnotationConfig` currently has two alphas:
- `alpha`: controls bounding boxes, attribute backgrounds, and instance segmentation borders
- `mask_fill_alpha`: controls the interior of instance segmentation masks

The default is `alpha = 0.75` and `mask_fill_alpha = 0.5` so that the border stands out on instance masks. It would be nice for confidence-scaling to preserve the two-levels of transparency here.

One can always set `alpha = 1` to recover the behavior of your implementation here.

tylerganter: good idea! done!

for your comment on the different "meaning" of confidence, I think this feature may help bring this to light. it is certainly possible to modify the confidence, for example:
```
confidence = confidence - threshold / (1 - threshold)
```

Regardless of what we do, having objects that fade out when they don't look good and are near 100% when they are clear (subjectively as a human) is a sign of a good model/system

PR #379, opened 2019-12-29
Events with Attributes comes to ETA!

jasoncorso:
This PR introduces events as a new parallel path to objects as far as content that ETA knows about in extraction and representation.

There are two main parts to this PR.
1. An enhancement to represent events in a similar fashion to objects in new `DetectedEvent` and `DetectedEventContainer` along with associated types, etc., and `VideoEventDetector` to actually process a video and extract events (abstract base class).
2. Changes to VideoLabels to represent events as part of the ETA label-space. It is implemented a "top-level" participant in the VideoLabels class much like video-level attributes, since the assumption is that no attributes of an event change at the frame level (the objects and frame-level attributes of the event are free to vary at the frame-level).  There is no explicit link between event and objects or attributes in this code.  This seemed cleanest to me while not imposing any backwards incompatibililies.

NOTE
- This PR builds atop the (outdated) `Event` and `EventSeries` classes in `eta.core.events`.  My proposal is to keep it this way for this PR and in a subsequent PR to remove `Event` and `EventSeries` and replace them with the `eta.core.video.FrameRange` and `eta.core.video.FrameRanges`.  But, I propose to keep the `EventDetection` class in `eta.core.events` but rename it to something more descriptive.  These changes will require some updates to other modules, etc., but this is minimal. 
- ETA needs an event detection pipeline/module to use this code in pipelines.  It is not part of this PR at this time.


eta/core/video.py
```diff
@@ -969,12 +1066,18 @@ def __init__(self, attrs=None, frames=None, objects=None):
             objects: a dictionary mapping object labels to
                 AttributeContainerSchema instances describing the object
                 attributes of each object class
+            events: a dictionary mapping event labels to
```
brimoor: this is going to get complicated if/when events can contain objects!

eta/core/events.py
```diff
@@ -27,7 +31,194 @@
...
+            event_series: the EventSeries (discontiguous temporal) range of the
+                event
+            confidence: (optional) the detection confidence in [0, 1]
+            index: (optional) an index assigned to the event
+            score: (optional) a score assigned to the event
+            attrs: (optional) an AttributeContainer describing additional
+                attributes of the event
+        '''
+        self.label = label
+        self.event_series = event_series
```
brimoor: I'm super motivated to replace `EventSeries` with `eta.core.video.FrameRanges` so that we can rename this field to `frames`.

A bit confusing that an "event" contains an "event_series" field.

I plan to make a PR into this that makes that tweak

eta/core/video.py
```diff
@@ -553,23 +554,30 @@ class VideoLabels(Serializable):
...
+    using VideoLabels in a dict-like fashion, frames are retrieved, not
+    attributes and not events.
+
     Attributes:
         filename: the filename of the video
         metadata: a VideoMetadata describing metadata about the video
         attrs: an AttributeContainer containing video attributes
         frames: a dictionary mapping frame number strings to VideoFrameLabels
             instances
+        events: a DetectedEventContainer containing events
```
brimoor: A common pattern for working with `VideoLabels` is:

```py
video_labels = etav.VideoLabels(...)

for frame_number in VideoLabels:
    frame_labels = video_labels[frame_number]  # eta.core.video.VideoFrameLabels
    # do stuff with the labels for the frame
```

This will not contain any attributes for events that coincide with a given frame.

Now that VideoLabels contains semantic things, we need to decide how to deal with frame-based iteration. Do we add support for "rendering" events for a given frame, which are stored in `VideoFrameLabels` "on-demand"?

We need a way to iterate over the frames of a VideoLabels that contains _everything_ for that frame. Player51 needs that, for example.

jasoncorso: My view, again, is that this not any more "semantic" than what there, but I think you are using that word for something different than I do.  If VideoLabels "needs" to remain only frame-based, then the proper thing to do is move events  to frames like objects.  However, I'll push back on this assumption "We need a way to iterate over the frames of a..."  Why?  Many representations require the rendered to ingest, visualize.  This is just a tacit assumption that has been made.

jasoncorso:
Representations are complicated.  After considering this at length, I think there are certain core assumptions about code / classes that are not sufficiently articulated.  So, this leads me to believe that, if for some reason, frame based access to all content in a video is core, then, this is an interface contract and representational contract.  That said, the way I believe all video data should be represented is in my PR from last year.  My thoughts have not changed much, although now I do think it important to distinguish between permanent and time varying attributes.

There are, however, many ways to represent this.  A second would be define all events and objects top level and then index into them over frame.  A third is strict frame-based as now.  

If the representational and interface contracts must match, then this PR should be refactored to make events frame-level things like objects.  (The sky may fall, but this would be the right way to meet those assumptions.)

brimoor:
I agree with what you're saying. This is the conversation I wanted to have so we can make the decisions and articulate the model.

I don't think that the representation must equal the interface, only that, since that happens to be the case before this PR, we should articulate new rules now.

Use cases that currently exist:
- depositing detections at the frame-level into VideoLabels
- iterating over all frame-based detections, for the purposes of reasoning about them (i.e., deciding to group some into a "trajectory")
- iterating over _everything_ in a frame, for the purposes of visualizing it
- iterating over the object "trajectories" in a video
- iterating over the events in a video, or using the event-membership of a given frame to inform processing of other information in that frame

Observation:In practice, since one can only have a few frames of a video in memory at a time, one often wants to work with the labels for a specified frame or small set of frames. So, an interface that gives access to that information is valuable. I don't care if this matches the representation, only that

mattphotonman:
> I agree with what you're saying. This is the conversation I wanted to have so we can make the decisions and articulate the model.
> 
> I don't think that the representation must equal the interface, only that, since that happens to be the case before this PR, we should articulate new rules now.
> 
> Use cases that currently exist:
> 
> * depositing detections at the frame-level into VideoLabels
> * iterating over all frame-based detections, for the purposes of reasoning about them (i.e., deciding to group some into a "trajectory")
> * iterating over _everything_ in a frame, for the purposes of visualizing it
> * iterating over the object "trajectories" in a video
> * iterating over the events in a video, or using the event-membership of a given frame to inform processing of other information in that frame
> 
> Observation:
> In practice, since one can only have a few frames of a video in memory at a time, one often wants to work with the labels for a specified frame or small set of frames. So, an interface that gives access to that information is valuable

One thought: what about storing the event information both at the top level in the `VideoLabels.events` attribute, as is done here, _and_ at the frame level.  This would significantly increase the space complexity, but would make it easier to access the event information in both frame-level and event-level fashion.  (In that case, user should only use the `VideoLabels` class methods to create and alter objects, in order to guarantee the consistency of the two internal representations.)

eta/core/events.py
```diff
@@ -27,7 +31,194 @@
...
+    def remove_events_without_attrs(self, labels=None):
+        '''Filters the events from this container that do not have attributes.
+
+        Args:
+            labels: an optional list of DetectedEvent label strings to which
+                to restrict attention when filtering. By default, all event
+                are processed
+        '''
+        filter_func = lambda event: (
+            (labels is not None and event.label not in labels)
```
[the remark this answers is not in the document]
benjaminpkane: I think this is correct as it is. The extra `labels` keyword is used to *restrict* the filter to certain labels. So objects without a label in `labels` are not targeted and therefore not filtered out.

These `lambda`s are pretty easy write, so if anything, I would say this function may not need to be a proper method for the class.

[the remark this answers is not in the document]
mattphotonman: If this is the case, then the doc string wording is a bit confusing.  I would call it a "black list" of labels to express what @benjaminpkane is saying.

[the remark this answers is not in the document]
brimoor: @mattphotonman it's implemented correctly and the docstring is correct. you're probably confused because you expect `filter_elements` to remove elements for which the filter is `True`, but it *keeps* elements them

eta/core/events.py
brimoor: done

PR #380, opened 2019-12-30
Adding support for exposing class probabilities for predictions/detections

brimoor:
Adding support for exposing class probabilities for predictions/detections

Release notes
- adding an `eta.core.learning.ExposesProbabilities` mixin to encapsulate models that expose class probabilities for their predictions/detections
- adding support for exposing probabilities for all detectors in `eta.detectors` and classifiers in `eta.classifiers`
- deleting `eta.core.learning.FeaturizingClassifier` and `eta.core.learning.FeaturizingDetector` in favor of a simpler `eta.core.learning.ExposesFeatures` interface. The relevant details of the different shapes of the features are relegated to the class docstrings
- adding support for saving all of the features/probabilities generated by the `detect_all` and `predict_all` methods
- adding support for recording top-k class probabilities to the relevant ETA modules


eta/classifiers/tfslim_classifiers.py
```diff
@@ -250,50 +253,126 @@ def features_dim(self):
...
+        attrs = etad.AttributeContainer()
+        keep = (confidence > self.config.confidence_thresh)
+        if keep:
+            attrs.add(self._package_attr(label, confidence))
+
+        return attrs, keep
+
+    def _package_attr(self, label, confidence):
+        return etad.CategoricalAttribute(
+            self.config.attr_name, label, confidence=confidence)
```
tylerganter: This whole function is obsolete

eta/core/learning.py
```diff
@@ -82,6 +84,18 @@ def write_labels_map(labels_map, outpath):
...
+def get_class_labels(labels_map):
+    '''Returns the list of class labels from the given labels map.
+
+    Args:
+        a dictionary mapping indexes to label strings
+
+    Returns:
+        a list of class labels
+    '''
+    return [labels_map[k] for k in sorted(labels_map.keys())]
```
tylerganter: What about circumstances where the labels map is 1-based instead of 0-based?

It seems like this function makes an assumption and as such should have a check that the keys are incrementing integers starting at 0

eta/classifiers/tfslim_classifiers.py
brimoor: what does obsolete mean? The fact that it's only 1 line long? It allows the caller to not wrangle `self.config.attr_name` themselves

tylerganter: I suppose...
But it's just moving a single line to another function, and it's only called in one place.
It could just as easily be:
```
packaged_attr = etad.CategoricalAttribute(
    self.config.attr_name, label, confidence=confidence)
attrs.add(packaged_attr)
```

eta/classifiers/vgg16_classifiers.py
```diff
@@ -111,43 +142,46 @@ def predict(self, img):
...
-            imgs: a list (or d x ny x nx x 3 tensor) of images to classify
+            imgs: a list (or n x h x w x 3 tensor) of images to classify
 
         Returns:
             a list of `eta.core.data.AttributeContainer` instances describing
                 the predictions for each image
         '''
+        return self._predict(imgs)
+
+    def _predict(self, imgs):
```
tylerganter: good call. I imagine these could speed things up

eta/classifiers/tfslim_classifiers.py
brimoor: Alright its gone

eta/classifiers/tfslim_classifiers.py
```diff
@@ -131,7 +131,7 @@ def __init__(self, d):
             d, "features_name", default=None)
         self.output_name = self.parse_string(d, "output_name", default=None)
         self.confidence_thresh = self.parse_number(
-            d, "confidence_thresh", default=None)
+            d, "confidence_thresh", default=0)
```
tylerganter: Module defaults should also be zero

eta/core/learning.py
brimoor: I updated this to ensure that all IDs between `min(1,min(labels_map))` and `max(labels_map)` will have labels generated. This handles cases where there are missing labels within the map, which, annoyingly, *IS* true for some COCO label maps (at one point the list was expanded from 80 to 91 labels).

0 vs 1-based indexing doesn't matter here- either will work. Its the user's responsibility to ensure that this list aligns with any vectors of class probabilities they generate

brimoor: yes, technically, if one is using 0-based indexing then `min(1,min(labels_map))` should be `min(0,min(labels_map))`, but its silly to provide a label map that doesn't contain the min or max label, so I'm not worried about that edge case

eta/modules/apply_classifier_to_objects.py
```diff
@@ -174,6 +174,8 @@ class ParametersConfig(Config):
             in pixels, for a bounding box to be classified
         confidence_threshold (eta.core.types.Number): [None] the minimum
             confidence required for a label to be saved
+        record_top_k_probs (eta.core.types.Number): [None] top-k class
```
mattphotonman: Slight clarification for the description: "top-k class probabilities to record for the predictions. (`record_top_k_probs` is the number `k`)."

eta/classifiers/tfslim_classifiers.py
```diff
@@ -177,13 +178,15 @@ def __init__(self, config):
...
+        # Load class labels
+        labels_map = etal.load_labels_map(self.config.labels_path)
+        self._class_labels = etal.get_class_labels(labels_map)
+        self._num_classes = len(self._class_labels)
 
         # Get network
         network_name = self.config.network_name
         network_fn = nets_factory.get_network_fn(
-            network_name, num_classes=len(self.labels_map), is_training=False)
+            network_name, num_classes=self._num_classes, is_training=False)
```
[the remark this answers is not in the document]
brimoor: Debatable. I don't like calling properties from inside the constructor. The object isn't even "built" yet!

eta/classifiers/tfslim_classifiers.py
```diff
@@ -250,50 +253,124 @@ def features_dim(self):
...
+            an array of features, or None if the classifier has not (or does
+                not) generate features
         '''
-        if not self.generates_features:
+        if not self.exposes_features:
             return None
 
         return self._last_features
 
+    def get_probabilities(self):
```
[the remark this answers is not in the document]
brimoor: I like it as implemented because it gives implementations the flexibility to store their probabilities however they choose. Maybe they choose to access them directly from a network graph, for example. Who knows

It is true that all of the implementations currently in our library follow a similar pattern of using `self._last_probs`, but it need not be the case

eta/modules/apply_classifier_to_objects.py
brimoor: addressed

PR #381, opened 2019-12-30
DetectedObject event indices (Proposal)

benjaminpkane:
This PR introduces object associations to events with an `event_indices` attribute on `DetectedObjects`.


benjaminpkane:
> @benjaminpkane can you submit this as a PR into `events2020` instead? I'd like to maintain that as a feature branch until questions about representation and interfaces to VideoLabels are settled. Thanks!

Done

eta/core/objects.py
```diff
@@ -107,7 +112,7 @@ def attributes(self):
         _attrs = ["label", "bounding_box"]
         _optional_attrs = [
             "mask", "confidence", "index", "score", "frame_number",
-            "index_in_frame", "eval_type"]
+            "index_in_frame", "eval_type", "event_indices"]
```
brimoor: As implemented, this will always include `event_indices` in every serialized `DetectedObject`. I'd rather only include it if the object has an associated event

benjaminpkane: Right, of course!

benjaminpkane:
> Key questions to resolve before merging:
> 
> * do we want to allow objects to belong to multiple events, or only zero/one?

It is fairly easy for me to think of cases where an object could be associated with two or more events, although they are all contrived at the moment. I see it as an easy win to have this flexibility.

PR #382, opened 2020-01-01
Adding two useful help-related features to the CLI

brimoor:
This PR adds two convenient features:

1. Running a naked `eta <command>` will now print help for that command. Previously, it would just execute the command, which may do something, do nothing, or give an error.

This is useful because a typical pattern for exploring a CLI is to:
- Run `eta` to see what the tool is all about (expecting to get help info)
- See an interesting command in the above list
- Run `eta <command>` to learn more about that command (expecting to get help info)

2. Automagically adding an `--all-help` flag for all commands that have subcommands. This is useful for seeing everything that one can do with a given command. Or, if one is so inclined, running `eta --all-help` to see EVERYTHING

### DIFF
--- a/eta/core/cli.py
+++ b/eta/core/cli.py
@@ -68,14 +68,15 @@ class Command(object):
         raise NotImplementedError("subclass must implement setup()")
 
     @staticmethod
-    def run(args):
-        '''Execute the command on the given args.
+    def run(parser, args):
+        '''Executes the command on the given args.
 
         args:
+            parser: the `argparse.ArgumentParser` instance for the command
             args: an `argparse.Namespace` instance containing the arguments
                 for the command
         '''
-        pass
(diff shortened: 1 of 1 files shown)


tylerganter:
Just tested. This only seems to be working for some commands.
`auth`, `http`, `s3` worked
`build`, `modules` did not

brimoor:
Clarification: this PR only addresses "meta" commands that are only composed of subcommands. Ie, `eta auth` is just a group of subcommands `eta auth <command>`, so this PR shows help for that.

But `eta build` is a leaf command that itself has flags, so it's not covered by this PR.

The ask is really to show help anytime an empty `argparse.Namespace` object is passed to `Command.run`. But I haven't found a clean way to do that. I'll make an issue for it.

brimoor:
https://github.com/voxel51/eta/issues/387

Issue #373, opened 2019-12-23
`eta.core.video` missing seek functionality on `VideoReader`s


jasoncorso:
This is a basic need that surprisingly has not arisen yet with the use of ETA, but there is no implemented support for video seeking.  The underlying functionality in FFmpeg and OpenCV readers should already include this, which would just need to be wrapped.

brimoor:
The way to seek now is via the `frames` argument of `*VideoReader`. But behind the scenes it reads frame-by-frame.

Since video readers support passing in arbitrary numbers of frame ranges of interest, the simplest optimization we could do is throw in an `-ss` flag to skip to the first frame/keyframe and then read from there.

brimoor:
Scope amended based on https://github.com/voxel51/eta/pull/383.

From @jasoncorso:
> VideoReader will need to support seeking formally in order for this to work as modules may need to go back and forth or at the very least "reopen" a video

brimoor:
Scope back to just `seek`, in light of https://github.com/voxel51/eta/pull/384.

PR #383, opened 2020-01-01
Passing VideoReader to video models

brimoor:
This PR updates the `eta.core.learning` interface for models that previously took paths to videos on disk to instead take `VideoReader` instances that can read the frames.

This is preferable because it separates the abstraction of loading the video from applying the model.

The impetus for this was the need to write modules that apply video models to _either_ a video path/frames sequence, which can be read by `eta.core.video.FFmpegVideoReader` _or_ a video frames directory containing not necessarily sequential frames, which can be read by `eta.core.video.SampledFramesVideoReader`. Much easier to do now!

Note that this is safe to do now because there are not yet any concrete instances of `VideoModel`, `VideoClassifier`, or `VideoObjectDetector` in `develop`.

### DIFF
--- a/eta/core/learning.py
+++ b/eta/core/learning.py
@@ -293,11 +293,12 @@ class VideoModel(Model):
     does not fit any of the concrete classifier/detector interfaces.
     '''
 
-    def process(self, video_path):
+    def process(self, video_reader):
         '''Generates labels for the given video.
 
         Args:
-            video_path: the path to the video
+            video_reader: an `eta.core.video.VideoReader` that can be used to
+                read the video
 
         Returns:
             an `eta.core.video.VideoLabels` instance containing the labels
@@ -472,11 +473,12 @@ class VideoClassifier(Classifier):
     that featurizes the frames of the input video.
     '''
 
-    def predict(self, video_path):
+    def predict(self, video_reader):
         '''Peforms prediction on the given video.
 
         Args:
-            video_path: the path to the video
+            video_reader: an `eta.core.video.VideoReader` that can be used to
+                read the video
 
         Returns:
             an `eta.core.data.AttributeContainer` instance containing the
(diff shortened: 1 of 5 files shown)


jasoncorso:
This is the right thing to do but two comments/questions:
1.  The pipeline system sees a video path and knows to create a video
reader?  What if, previously, certain other module parameters would impact
how the video was read, such as ranges and rates?
2.  VideoReader will need to support seeking formally in order for this to
work as modules may need to go back and forth or at the very least "reopen"
a video.

jasoncorso:
This is the right thing to do but two comments/questions:
1.  The pipeline system sees a video path and knows to create a video reader?  What if, previously, certain other module parameters would impact how the video was read, such as ranges and rates?
2.  VideoReader will need to support seeking formally in order for this to work as modules may need to go back and forth or at the very least "reopen" a video.

brimoor:
1. This module shows how one would configure the VideoReader as desired (any sources, any parameters, etc): https://github.com/voxel51/eta/pull/383/files#diff-73bfe1f6c2c139a1be5f0b3b743ccacd

2. Yes good point. I'll leave the work of upgrading VideoReader to expressly support seeking/re-reading videos to the first person who needs that functionality. Shouldn't be too hard. https://github.com/voxel51/eta/issues/373 is updated to reflect this ask

eta/core/learning.py
```diff
@@ -472,11 +473,12 @@ class VideoClassifier(Classifier):
     that featurizes the frames of the input video.
     '''
 
-    def predict(self, video_path):
+    def predict(self, video_reader):
```
mattphotonman: Isn't there some code in the concrete subclasses that needs to be modified accordingly?  Or are there currently no concrete subclasses of `VideoClassifier`?

benjaminpkane: There are no concrete subclasses.

PR #389, opened 2020-01-04
Apply Classifier to `LabeledImageDataset`

tylerganter:
Module can now take a dataset as input (very useful when you don't want to process all images in a directory!)

~~I just need to test it~~
Tested and works beautifully!


eta/modules/apply_image_classifier.py
```diff
@@ -144,6 +150,18 @@ def _validate(self):
                     "`output_image_set_labels_path` is required when "
                     "`images_dir` is set")
 
+        if self.image_dataset_path:
+            if not self.output_image_set_labels_path:
+                raise ConfigError(
```
mattphotonman: I think this should be a `ValueError`. `ConfigError` looks like it's more for basic `Config` inconsistencies, like when a variable has no default and a value isn't provided by the user.

tylerganter: I'm just following the convention of the code right above this. But for me...GS0. I think they do the same thing

mattphotonman: Oh you're right, that's fine.  It only matters if you're using `try` / `except` statements that are looking for specific exception types.

PR #392, opened 2020-01-07
Bounding Box corner functionality

benjaminpkane:
This PR adds convenience properties `top_right`, `bottom_right` and `corners`.

### DIFF
--- a/eta/core/geometry.py
+++ b/eta/core/geometry.py
@@ -67,11 +67,12 @@ class BoundingBox(Serializable):
 
     @property
     def corners(self):
-        '''Returns all four RelativePoint corners in clockwise order starting with
-        the top left corner.
-        
+        '''Returns all four RelativePoint corners in clockwise order starting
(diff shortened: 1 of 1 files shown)


eta/core/geometry.py
```diff
@@ -55,6 +55,27 @@ def __eq__(self, bbox):
...
+        '''Returns a bottom left RelativePoint'''
+        return RelativePoint(self.top_left.x, self.bottom_right.y)
+
+    @property
+    def corners(self):
+        '''Returns all four RelativePoint corners in clockwise order starting with
+        the top left corner.
+        
+        Returns:
+            a 4-tuple of RelativePoints
```
brimoor: small suggestion: could write this as `(top_left, top_right, bottom_right, bottom_left) tuple of RelativePoints` to make it easier on the reader of the documentation only

benjaminpkane: Done!

PR #395, opened 2020-01-07
adding LabeledFileRecord to ETA

jasoncorso:
This is a useful and reusable class for working with low-level datasets.  Adding it to `core.data`


jasoncorso:
I will make this improvement later on today.  I think maintaining a low-level code-base beneath `LabeledDataset` is not a bad idea.  One often wants to wrangle from scratch....  That was the birth of all of these `DataRecords` classes.

PR #396, opened 2020-01-07
Option to relabel object indices before merging `VideoLabels`

mattphotonman:
The `reindex_objs` option allows you to reindex the object indices so that there will not be any conflicting indices once a new `VideoLabels` is merged into the existing one.  This option is `False` by default, for back-compatibility.


eta/core/video.py
```diff
@@ -738,12 +738,19 @@ def get_frame_range(self):
...
 
         Args:
             video_labels: a VideoLabels instance
+            reindex_objs: whether to reindex objects in `self` before merging
+                so that all indices are different than indices in
+                `video_labels`
         '''
+        if reindex_objs:
+            self._reindex_objs(video_labels, "index")
+            self._reindex_objs(video_labels, "index_in_frame")
```
brimoor: `index_in_frame` means the position of the `DetectedObject` in the container in its associated frame. It's used to associated bounding boxes with, for example, features that are separately written to disk. So this isn't doing the right thing unfortunately

eta/core/video.py
```diff
@@ -738,12 +738,19 @@ def get_frame_range(self):
         fns = self.get_frame_numbers()
         return (fns[0], fns[-1]) if fns else (None, None)
 
-    def merge_video_labels(self, video_labels):
+    def merge_video_labels(self, video_labels, reindex_objs=False):
         '''Merges the given VideoLabels into this labels.
 
         Args:
             video_labels: a VideoLabels instance
+            reindex_objs: whether to reindex objects in `self` before merging
```
brimoor: small thing, but it seems the more natural thing to do would be to change the `index`es of the *new objects*, not the ones in `self`

eta/core/video.py
```diff
@@ -962,6 +969,45 @@ def _validate_schema(self):
...
+        '''Reindexes objects in `self` so that all indices are different than
+        indices in the input.
+
+        Args:
+            video_labels: a VideoLabels instance
+            variable_name: the name of the member variable of `DetectedObject`
+                that contains the index. The value of this variable is assumed
+                to be an integer.
+        '''
+        self_object_indices = [
```
brimoor: would be much more efficient to implement this as a `set()`

eta/core/video.py
brimoor: Also, this implies that `index_in_frame` cannot be changed in a vacuum.

Or, rather, one really shouldn't be merging `VideoLabels` whose objects have `index_in_frame` if one actually cares about _using_ those values. I take the blame for this because the meaning of the field is not very clearly documented

eta/core/video.py
mattphotonman: but that would mutate the input `VideoLabels`, whereas `self` is going to be changed internally anyways due to merging.

mattphotonman:
@brimoor ready for re-review.  I changed it back to reindexing the indices of `self` instead of the input `video_labels`.

PR #398, opened 2020-01-07
Tweaking implementation of reindexing objects

brimoor:
@mattphotonman here's a tweaked implementation of reindexing


brimoor:
Don't worry about `index_in_frame` no longer being reindexed. Changes I'm working on will likely remove that field anyhow

eta/core/video.py
```diff
@@ -738,18 +738,16 @@ def get_frame_range(self):
...
-    def merge_video_labels(self, video_labels, reindex_objs=False):
+    def merge_video_labels(self, video_labels, reindex=False):
         '''Merges the given VideoLabels into this labels.
 
         Args:
             video_labels: a VideoLabels instance
-            reindex_objs: whether to reindex objects in `self` before merging
-                so that all indices are different than indices in
-                `video_labels`
+            reindex: whether to reindex objects in `video_labels` before
```
brimoor: I renamed this parameter to `reindex` because https://github.com/voxel51/eta/pull/379 is adding events, which will also need to be reindexed here via the same method

PR #399, opened 2020-01-08
Attribute Container accessor helpers

tylerganter:
I think y'all will be happy with this :)


eta/core/data.py
```diff
@@ -531,6 +532,56 @@ def remove_schema(self):
         '''Removes the enforced schema from the container.'''
         self.schema = None
 
+    def get_attrs_with_name(self, name):
+        '''Get a list of all attributes with a given name
+
+        Args:
+            name: the Attribute name
+        Returns:
```
benjaminpkane: nit: new line after indentation

eta/core/data.py
```diff
@@ -531,6 +532,56 @@ def remove_schema(self):
...
+            a list of attributes with the given name
+        '''
+        return [attr for attr in self if attr.name == name]
```
mattphotonman: Why not return an `AttributeContainer`?

eta/core/data.py
```diff
@@ -531,6 +532,56 @@ def remove_schema(self):
...
+        Args:
+            name: the Attribute name
+        Returns:
+            the Attribute
+        Raises:
+            ValueError if there is not exactly one Attribute with the name
+            `name`
+        '''
+        attrs = self.get_attrs_with_name(name)
+        if len(attrs) != 1:
```
mattphotonman: What do you think about adding a statement
```python
if len(attrs) == 0:
    return None
```
before this line.

tylerganter: Well then there's not distinguishing the attr value if the value is `None`, right?

eta/core/data.py
```diff
@@ -531,6 +532,78 @@ def remove_schema(self):
...
+        Args:
+            name: the Attribute name
+
+        Returns:
+            an AttributeContainer of attributes with the given name
+        '''
+        kwargs = {
+            self._ELE_ATTR: [attr for attr in self if attr.name == name]
+        }
+        return self.__class__(**kwargs)
```
brimoor:
```py
return self.get_matches([lambda attr: attr.name == name])
```

eta/core/data.py
```diff
@@ -531,6 +532,78 @@ def remove_schema(self):
...
+        give name.
+
+        Args:
+            name: the Attribute name
+
+        Returns:
+            boolean indicating whether or not the container contains an
+                Attribute with the given name
+        '''
+        return bool(self.get_attrs_with_name(name))
```
brimoor: It would be more efficient for large containers to do:
```py
for attr in self:
    if attr.name == name:
        return True

return False
```

eta/core/data.py
```diff
@@ -531,6 +532,78 @@ def remove_schema(self):
...
+
+    def get_attrs_with_name(self, name):
+        '''Get a list of all attributes with a given name
```
brimoor: prefer not to say "list" here because you're actually returning a `Container`.

Really glad that the suggestion was already made to return a `Container`, not a `list`. `list`s of `Attribute`s should never be created. `AttributeContainer` is a fancy way to store lists of `Attribute`s that provides more convenience methods

eta/core/data.py
mattphotonman: I got this one, since I added this method.

mattphotonman: Fixed.

eta/core/data.py
mattphotonman: Made a PR with a change that should make us both happy: #401

PR #402, opened 2020-01-10
Fix dataset splits

benjaminpkane:
I believe `dataset_index` was made a property at some point. The `split` needs to set `_dataset_index` now because of this.

### DIFF
--- a/eta/core/datasets.py
+++ b/eta/core/datasets.py
@@ -598,6 +598,20 @@ class LabeledDataset(object):
         '''
         return self._dataset_index
 
+    @dataset_index.setter
+    def dataset_index(self, dataset_index):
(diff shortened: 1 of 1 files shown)


eta/core/datasets.py
```diff
@@ -744,7 +744,7 @@ def split(self, split_fractions=None, descriptions=None,
         dataset_copy = copy.deepcopy(self)
         dataset_list = []
         for dataset_index in dataset_indices:
-            dataset_copy.dataset_index = dataset_index
+            dataset_copy._dataset_index = dataset_index
```
mattphotonman: Better to create a setter method for `dataset_index`.

benjaminpkane: Done!

benjaminpkane:
@mattphotonman Ready for review

PR #405, opened 2020-01-16
Image/VideoSetLabels from glob pattern

tylerganter:
Any instance of `Set` that has an element class that is serializable may be loaded from a numeric or glob element pattern. I also deprecated the class methods in `eta.core.{image, video}`

Example usage:

```
set_labels = etav.VideoSetLabels.from_numeric_patt("%d_tele.json")
set_labels2 = etav.VideoSetLabels.from_glob_patt("*.json")
```

### DIFF
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -834,9 +834,8 @@ class ImageSetLabels(Set):
             a `cls` instance
         '''
         logger.warning("Using deprecated method `from_image_labels_patt`. Use"
-                       " `from_element_patt` instead.")
-        return cls.from_element_patt(
-            image_labels_patt, pattern_type=etas.NUMERIC)
+                       " `from_element_numeric_patt` instead.")
+        return cls.from_element_numeric_patt(image_labels_patt)
 
     @classmethod
     def from_dict(cls, d):
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -834,7 +834,7 @@ class ImageSetLabels(Set):
             a `cls` instance
         '''
         logger.warning("Using deprecated method `from_image_labels_patt`. Use"
-                       " `from_element_numeric_patt` instead.")
+                       " `from_numeric_patt` instead.")
         return cls.from_numeric_patt(image_labels_patt)
 
     @classmethod
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -835,7 +835,7 @@ class ImageSetLabels(Set):
         '''
         logger.warning("Using deprecated method `from_image_labels_patt`. Use"
                        " `from_element_numeric_patt` instead.")
-        return cls.from_element_numeric_patt(image_labels_patt)
+        return cls.from_numeric_patt(image_labels_patt)
 
     @classmethod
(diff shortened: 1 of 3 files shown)


eta/core/image.py
```diff
@@ -833,6 +834,22 @@ def from_image_labels_patt(cls, image_labels_patt):
             image_set_labels.add(cls._ELE_CLS.from_json(labels_path))
         return image_set_labels
 
+    @classmethod
```
mattphotonman: Why not just make this a method of the underlying `Set` class and call it `Set.from_glob_patt()`?  That way we don't have to re-write it for every `Set` subclass.

tylerganter: Great point!

eta/core/serial.py
```diff
@@ -824,6 +835,36 @@ def from_dict(cls, d, **kwargs):
         elements = [cls._ELE_CLS.from_dict(dd) for dd in d[cls._ELE_ATTR]]
         return cls(**etau.join_dicts({cls._ELE_ATTR: elements}, kwargs))
 
+    @classmethod
+    def from_element_patt(cls, pattern, pattern_type=NUMERIC):
```
mattphotonman: I suggest adding arguments `*args, **kwargs` that will be passed to the `from_json()` method.  See [here](https://github.com/voxel51/eta/blob/develop/eta/core/serial.py#L388).

tylerganter: happy now?!

eta/core/serial.py
```diff
@@ -49,6 +50,7 @@
 logger = logging.getLogger(__name__)
 
 
+
```
brimoor: unnecessary newline

eta/core/serial.py
```diff
@@ -824,6 +826,47 @@ def from_dict(cls, d, **kwargs):
...
 
+    @classmethod
+    def from_numeric_patt(cls, pattern, *args, **kwargs):
+        '''Creates an instance of `cls` from a numeric pattern of `_ELE_CLS`
+        files.
+
+        Args:
+             pattern: a pattern with one or more numeric sequences
+                example: "/path/to/labels/%05d.json"
+            *args: optional positional arguments for `cls.from_json()`
```
brimoor: `_ELE_CLS.from_json`, not `cls.from_json`

eta/core/video.py
```diff
@@ -1931,10 +1932,9 @@ def from_video_labels_patt(cls, video_labels_patt):
         Returns:
             a `cls` instance
         '''
-        image_set_labels = cls()
-        for labels_path in etau.get_pattern_matches(video_labels_patt):
-            image_set_labels.add(cls._ELE_CLS.from_json(labels_path))
-        return image_set_labels
+        logger.warning("Using deprecated method `from_video_labels_patt`. Use"
+                       " `from_element_numeric_patt` instead.")
+        return cls.from_numeric_patt(video_labels_patt)
```
brimoor: `from_element_numeric_patt` is incorrect in this warning message

PR #408, opened 2020-01-17
Enhancements to annotation rendering via ETA

brimoor:
The following new parameters have been added to `eta.core.annotations.AnnotationConfig` (and thereby usable when running the `visualize_labels` module or `visualize_labels` pipeline):

The only non-backwards compatible change is that all object attributes will now be rendered as a `panel`.

**Object attribute rendering**
- `object_attrs_render_method = ["list", "panel"]`: the method used to render object attributes. Previous behavior was `list`. New option is `panel`, which renders object attributes in a panel in the same way that frame attributes are rendered. I find this far easier to grok, so I made `panel` the new default

**Handling of occluded objects**
- `occluded_object_attr`: the name of the boolean attribute indicating whether an object is occluded. Typically this is `occluded`
- `hide_occluded_objects = True/False`: whether to hide objects when their `occluded_object_attr` takes the value `False`

**Handling of boolean attributes**
- `hide_false_boolean_attrs = True/False`: whether to hide boolean attributes when they are `False`, rather than rendering `<attr>: False`

**Hiding of specific attribute values**
- `hide_attr_values = <list>`: an optional list of video/frame/object attribute values to NOT RENDER if they appear in the labels. A typical use of this is to hide attributes whose value is `unclear`

Example of how object attributes now look in panel-mode (the new default):
Screen Shot 2020-01-17 at 1 36 21 PM

Same labels as above, but hiding `unclear` attributes:
Screen Shot 2020-01-17 at 1 37 02 PM

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -20,7 +20,6 @@ from builtins import *
 
 import logging
 import random
-import sys
 
 import cv2
 import numpy as np
@@ -29,6 +28,7 @@ from PIL import Image, ImageDraw, ImageFont
 import eta
 import eta.constants as etac
 from eta.core.config import Config, Configurable
+import eta.core.data as etad
 import eta.core.image as etai
 import eta.core.logo as etal
 import eta.core.video as etav
@@ -50,7 +50,7 @@ class AnnotationConfig(Config):
             available
         show_object_attr_confidences: whether to show object attribute
             confidences, if available
-        show_frame_attr_confidences: whether to show frame/video attribute
+        show_frame_attr_confidences: whether to show video/frame attribute
             confidences, if available
         show_all_confidences: whether to show all confidences, if available.
             If set to `True`, the other confidence-rendering flags are ignored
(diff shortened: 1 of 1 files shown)


eta/core/annotations.py
```diff
@@ -105,10 +124,14 @@ def __init__(self, d):
         self.confidence_scaled_alpha = self.parse_bool(
             d, "confidence_scaled_alpha", default=False)
         self.text_color = self.parse_string(d, "text_color", default="#FFFFFF")
+        self.object_attrs_render_method = self.parse_categorical(
+            d, "object_attrs_render_method", ["list", "panel"], default="list")
```
tylerganter: This is still default `list`

eta/core/annotations.py
```diff
@@ -581,14 +675,16 @@ def _render_attr_name_value(attr, show_confidence=True):
...
+    add_index = show_index and obj.index is not None
+
     label_str = _clean(obj.label).upper()
 
-    add_confidence = show_confidence and obj.confidence is not None
     if add_confidence:
         label_str += " (%.2f)"
 
-    if obj.index is not None:
+    if add_index:
```
tylerganter: `if add_index and obj.index is not None`

eta/core/annotations.py
brimoor: nice catch; thanks

eta/core/annotations.py
brimoor: `add_index` already includes that check

PR #414, opened 2020-01-28
initial draft of schema differencing

iantimmis:
[DRAFT: This feels done, but haven't tested yet. Doing that now]

A new function `diff_schema` has been as a method of comparing schemas. It has been added to the following classes:

AttributeSchemaContainer
AttributeSchema
BooleanAttributeSchema
NumericAttributeSchema
CategoricalAttributeSchema
ImageLabelsSchema
VideoLabelsSchema

(Anywhere where we find merge_schema)

Example:
```py
# Constructs a schema containing the set difference between schema_1 and schema_2
diff_set = schema_1.diff_schema(schema_2)

# Constructs schemas for attributes only contained in schema_1  and in schema_2 respectively
schema_1_only = schema_1.diff_schema(diff_set)
schema_2_only = schema_2.diff_schema(diff_set)

# If no difference is found, then None is returned
empty_schema = schema_1.diff_schema(schema_1)
```

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -275,6 +275,11 @@ class AttributeSchema(Serializable):
         '''Incorporates the given AttributeSchema into the schema.'''
         raise NotImplementedError("subclass must implement merge_schema()")
 
+    def diff_schema(self, schema):
+        '''Constructs a schema containing the set difference of the source and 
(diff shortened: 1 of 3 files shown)


tylerganter:
I think we can close this

PR #422, opened 2020-02-03
Labels infrastructure overhaul

brimoor:
**Release Notes**
- introducing a base `eta.core.labels.Labels` class from which all labels derive
- introducing a base `eta.core.labels.LabelsSchema` from which all labels schemas derive
- introducing a base `eta.core.labels.LabelsFrameRenderer` class that encapsulates the action of rendering information about spatiotemporal entities for an individual frame
- adding a base `eta.core.frames.FrameLabels` class to encapsulate `Labels` classes that describe content in individual images/video frames
- adding a new `eta.core.objects.Object` class that represents a spatiotemporal object in a video and all of its per-frame observations
- adding support for frame-level labels in `eta.core.events.Event`s 
- adding a new `eta.core.events.DetectedEvent` class to encapsulate per-frame information about an event
- adding support for `eta.core.events.Event`s to contain `Object`s and other `Event`s
- adding support for rendering `Object`s, `Event`s, and `VideoLabels` at the frame-level via the `LabelsFrameRenderer` interface

**Important Comments**
- EVERYTHING related to `Labels` in this PR is backwards compatible. All existing `VideoLabels` are still valid `VideoLabels`, for example
- The format of schemas has changed slightly. Previously serialized schemas will no longer be readable by their new `LabelsSchema` implementations

**TODO**
- add support for rendering ImageLabels/VideoLabels via `eta.core.annotations`
- add support for converting framewise Object/Event/VideoLabels to spatiotemporal format
- finish implementing schema validation for objects/events with children

**Test Code**

```py
from copy import deepcopy

import eta.core.data as etad
import eta.core.image as etai
import eta.core.objects as etao
import eta.core.video as etav

import theta.core.objects as theo

'''
# Download some test data to work with
eta gdrive download 129Vw0kkVY9eOBscdMcBLswx1ov6rXae6 /path/to/animals.json
eta gdrive download 1T82D1RUfRqpXKtzYwxBeEGyBXUIwpqLU /path/to/vehicles.json
'''

video_labels_path = "/path/to/vehicles.json"
image_set_labels_path = "/path/to/animals.json"

video_labels = etav.VideoLabels.from_json(video_labels_path)

schema = video_labels.get_active_schema()
print(schema)

#
# Convert DetectedObjects to Objects with object-level attributes
#

video_labels2 = etav.VideoLabels()

for traj in theo.TrajectorySet.from_video_labels(video_labels):
    obj = etao.VideoObject()

    all_attrs = etad.AttributeContainer()
[...]

PR #423, opened 2020-02-03
Dataset Cleaning Tools and Support

tylerganter:
This branch now just supports and old way of representing constant and exclusive attributes which will is not be needed one #422 is merged.

Once dataset cleaning supports the new video labels represntation this can be closed


brimoor:
https://github.com/voxel51/eta/pull/422 is merged so this is no longer needed

PR #428, opened 2020-02-11
Serialize boolean arrays more space-efficiently

lethosor:
`numpy.save()` saves these as uint8 arrays, which wastes 7/8 bits
* Followup to #426
* Open to changes that would improve readability of this
* Taking base64 encoding into account, this seems to result in space savings of between 3x and 4x for a 33x33 array (savings for larger arrays will be more significant)

### DIFF
--- a/eta/core/serial.py
+++ b/eta/core/serial.py
@@ -197,7 +197,18 @@ def serialize_numpy_array(array):
     # We currently serialize in numpy format. Other alternatives considered
     # were `pickle.dumps(array)` and HDF5
     #
+    header = None
+    if array.dtype == bool:
+        # pack boolean arrays more efficiently - this destroys size
+        # information, so we store it separately
+        header = b"bool" + np.array(
+            (len(array.shape),) + array.shape, dtype="uint32").tobytes()
+        array = np.packbits(array)
+
     with io.BytesIO() as f:
+        if header is not None:
+            f.write(header)
+
         np.save(f, array, allow_pickle=False)
         bytes_str = f.getvalue()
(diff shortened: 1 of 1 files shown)


eta/core/serial.py
```diff
@@ -197,7 +197,18 @@ def serialize_numpy_array(array):
     # We currently serialize in numpy format. Other alternatives considered
     # were `pickle.dumps(array)` and HDF5
     #
+    header = None
+    if array.dtype == bool:
```
brimoor: What about compressing the serialized bytes using something off-the-shelf?

The concern with introducing custom code here is that it will add extra work to any client who wants to support deserializing these masks for use in their code

brimoor (changes requested):

eta/core/serial.py
lethosor: I'll investigate zlib or something similar as well; this is guaranteed to achieve 8x reduction for arbitrary bit arrays, though (I don't think zlib can achieve that, and introducing zlib on the client side is considerably more complex, while this has already been implemented there)

brimoor: https://github.com/nodeca/pako?

lethosor: Yeah, I think that will work, but we'll need to investigate client-side performance. From initial tests, compressing a single large (1920x1080) mask takes around 150ms - decompressing may be faster.

brimoor: Let's get the system working and then think about implications of CPU vs bandwidth vs interface complexity

PR #429, opened 2020-02-12
Added subschema validation

iantimmis:
Added functionality to determine whether a VideoLabelsSchema is a subset of another VideoLabelsSchema. Likewise with ImageLabelsSchema.

```py
from eta.core.video import VideoLabelsSchema

ont_schema = VideoLabelsSchema.from_json('ontology.json')
cov_schema = VideoLabelsSchema.from_json('coverage.json')

print(ont_schema.is_valid_subset_schema(cov_schema)) # Prints True
```

### DIFF
--- a/eta/core/image.py
+++ b/eta/core/image.py
@@ -7,11 +7,12 @@ Notes:
         produced outside of this library must be converted to RGB. This
         conversion can be done via `eta.core.image.bgr_to_rgb()`
 
-Copyright 2017-2019, Voxel51, Inc.
+Copyright 2017-2020, Voxel51, Inc.
 voxel51.com
 
 Brian Moore, brian@voxel51.com
 Jason Corso, jason@voxel51.com
+Ian Timmis, ian@voxel51.com
 '''
 # pragma pylint: disable=redefined-builtin
 # pragma pylint: disable=unused-wildcard-import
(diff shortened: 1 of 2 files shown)


eta/core/image.py
```diff
@@ -485,6 +486,32 @@ def is_valid_object(self, obj):
         except (ImageLabelsSchemaError, AttributeContainerSchemaError):
             return False
 
+    def is_valid_subset_schema(self, schema):
+        '''Validates that the provided schema is a subset of self
+        
+        Args:
+            schema: an ImageLabelsSchema
+        Returns:
```
brimoor: one space between `Args:` and `Returns:` sections

eta/core/image.py
```diff
@@ -485,6 +486,32 @@ def is_valid_object(self, obj):
...
+            True/False
+        '''
+
+        # Validate frame attributes
+        for attr in schema.attrs:
+            if not self.is_valid_image_attribute(attr):
```
brimoor: This doesn't work. `attr` is an `AttributeSchema`, but `is_valid_image_attribute` expects an `Attribute`

eta/core/image.py
```diff
@@ -485,6 +486,32 @@ def is_valid_object(self, obj):
...
+                return False
+
+        # Validate objects
+        for obj in schema.objects:
+            if not self.is_valid_object_label(obj):
+                return False
+
+            # Validate object attributes
+            for attr in schema.objecs[obj].schema
```
brimoor: typo: `objecs` -> `objects`

eta/core/image.py
brimoor: The same comment applies throughout this PR. No functionality yet exists for, e.g., checking whether the `categories` of one `CategoricalAttributeSchema` are a subset of another. That will have to be written

iantimmis: Ah, gotcha. Thanks for the review. I'll work it out.

brimoor:
@iantimmis you can drop this PR. I implemented it in https://github.com/voxel51/eta/pull/422.

In case you're interested, here are the relevant commits:
https://github.com/voxel51/eta/pull/422/commits/9b22e792cb4180f25d4b359ce8d6ed9e5fd25d37
https://github.com/voxel51/eta/pull/422/commits/bccf396eebb04047a37bd30a1a4727bc2ffbd0ed
https://github.com/voxel51/eta/pull/422/commits/780177e38619b1ac290201cf1444b9d51fdec494
https://github.com/voxel51/eta/pull/422/commits/42654d692137bb1d94014a3abb34ecba7b5a312d

PR #433, opened 2020-02-20
Adding support for visualizing frame masks and events in images/videos

brimoor:
This PR builds on top of https://github.com/voxel51/eta/pull/422.

This PR adds support for visualizing all types of content that VideoLabels 2.0 can hold.

High-level features:
- automatic rendering of any video-level content in `VideoLabels` to the frame-level
- support for rendering frame masks
- support for rendering events

When rendering frame masks, semantic labels can optionally be added if `mask_index` fields are provided in the `VideoLabels`. The location of these semantic labels is set, per frame, at the centroid of the largest region taking each index value.

When rendering events, the following options are supported:
- option to render children of events in the same color as the event itself (assuming the event has a bounding box and/or instance mask)
- if the event does not have a bounding box, any event-level attributes are appended to the frame attribute panel in the upper-left
- option to render the label + index of an event as an attribute in the attribute panels of any child objects

`eta.core.annotations.AnnotationConfig` now contains ~40 parameters that enable you to customize how the labels are rendered.

Example of rendering instance masks:
example3

Example of rendering frame masks:
image-anno

Example of rendering an event with child objects:
(The filled bounding box was added just for fun; note that the child objects contain an `event` attribute that encodes the fact that they belong to the event)
example2

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -107,6 +107,8 @@ class AnnotationConfig(Config):
         mask_border_thickness: the thickness, in pixels, to use when drawing
             the borders of segmentation masks
         mask_fill_alpha: the transparency of segmentation masks
+        show_frame_mask_semantics: whether to render semantic labels for frame
+            mask regions, when available
 
     ##### ATTRIBUTE BOXES #####
         attrs_box_render_method: the method used to render object attributes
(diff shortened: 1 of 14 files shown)


eta/core/annotations.py
```diff
@@ -411,250 +532,543 @@ def annotate_image(img, image_labels, annotation_config=None):
...
+    c = max(contours, key=cv2.contourArea)
+
+    M = cv2.moments(c)
+    tlx = int(M["m10"] / M["m00"])
+    tly = int(M["m01"] / M["m00"])
+    return [(tlx, tly)]
 
-    return np.asarray(img_pil)
+
+'''
```
benjaminpkane: Does this need to be here?

brimoor: oops! removed

PR #435, opened 2020-02-20
Small improvements with a vague description

tylerganter:
- Sorting categorical schema categories such that the schema JSON is predictable when serialized
- exposing read/write labels for `LabeledDataset`
- verbose when finding duplicates (this takes  a long time generally)

### DIFF
--- a/eta/core/utils.py
+++ b/eta/core/utils.py
@@ -2257,12 +2257,13 @@ def remove_none_values(d):
     return {k: v for k, v in iteritems(d) if v is not None}
 
 
-def find_duplicate_files(path_list):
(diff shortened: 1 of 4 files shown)


eta/core/utils.py
```diff
@@ -2257,26 +2257,34 @@ def remove_none_values(d):
     return {k: v for k, v in iteritems(d) if v is not None}
 
 
-def find_duplicate_files(path_list):
+def find_duplicate_files(path_list, verbose=True):
```
brimoor: typical convention would be to set `verbose=False` by default, to avoid spam

PR #438, opened 2020-02-24
`AttributeContainerSchema.merge_attribute_schema`

tylerganter:
I need to be able to add a single `AttributeSchema` instance into an `AttributeContainerSchema`


tylerganter:
@brimoor good point, done

PR #448, opened 2020-03-04
Extract clip frames from video

brimoor:
Adding a method to extract the frames in a specified clip of a video.

Test code:

```py
'''
# Download test video to work with
eta gdrive download 1izpoUwMDPsmJMbyKzTgGmkoSgBk0a2IQ /path/to/test.mov
'''

import eta.core.video as etav

video_path = "/path/to/test.mov"

# Output frames with numbers 1, 2, 3, ...
output_patt1 = "/path/for/frames1/%05d.png"
etav.extract_clip_frames(
    video_path, output_patt1, start_time=1, duration=1,
    keep_source_frame_numbers=False)

# Output frames with their original frame numbers
output_patt2 = "/path/for/frames2/%05d.png"
etav.extract_clip_frames(
    video_path, output_patt2, start_time=1, duration=1,
    keep_source_frame_numbers=True)
```


eta/core/video.py
```diff
@@ -2211,7 +2245,9 @@ def extract_clip(
 
     Args:
         video_path: the path to the video
-        output_path: the path to write the extracted video clip
+        output_path: the path to write the extracted video clip. If
+            `fast == True`, this path can optionally be a pattern like
```
benjaminpkane: It is good to know that readers of the documentation will know this is possible. I would advocate for throwing an error, though, as it is not the methods intended use.

With all that said, I am assuming this was not changed for backwards compatibility.

eta/core/video.py
```diff
@@ -2260,9 +2298,51 @@ def extract_clip(
...
+    '''Extracts the frames of the specified clip from the video.
+
+    When `keep_source_frame_numbers` is False, the following ffmpeg command is
+    used to perform the sampling:
+    ```
+    ffmpeg -ss <start_time> -i <video_path> -t <duration> <output_path>
+    ```
+
+    When `keep_source_frame_numbers` is True, the following ffmpeg command is
+    used to perform the
```
benjaminpkane: perform the sampling*

PR #452, opened 2020-03-12
Turning off JSON pretty-printing by default when serializing

brimoor:
By popular demand, all `Serializable` objects will now be written to disk with no whitespace.

`Serializable.to_str()` still pretty-prints by default, however, so that `print(obj)` will be pretty.


brimoor:
We're not yet using black in ETA

PR #455, opened 2020-03-16
Module Metadata Bugfix

tylerganter:
Required video input made it impossible to run on images


tylerganter:
> The proper way to fix this is to add `[None]` to the docstrings for the corresponding inputs in the module `.py` file. Then re-run `eta modules --metadata path/to/eta/modules/apply_classifier_to_objects.py`

Right, my bad. Fixed!

eta/core/metadata.py
```diff
@@ -85,7 +85,7 @@ def generate(module_py_path):
 
     outpath = os.path.join(module_dir, module_name + ".json")
     logger.info("Writing module metadata to %s", outpath)
-    mmc.write_json(outpath)
+    mmc.write_json(outpath, pretty_print=True)
```
brimoor: thanks for adding this!

PR #456, opened 2020-03-17
Adding support for default values to AttributeSchema

brimoor:
`eta.core.data.AttributeSchema`s now support an optional `default` field that specifies the default value for the attribute in the schema.

This functionality is not currently _used_ anywhere in ETA, but is nonetheless useful in general when working with schemas.

### DIFF
--- a/eta/core/data.py
+++ b/eta/core/data.py
@@ -339,7 +339,7 @@ class AttributeSchema(etal.LabelsSchema):
         self.add(attr)
 
     def is_valid_value(self, value):
-        '''Whether value is valid for the attribute.
+        '''Whether the attribute value is compliant with the schema.
 
         Args:
             value: the value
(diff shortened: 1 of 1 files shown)


eta/core/data.py
```diff
@@ -293,26 +293,35 @@ class AttributeSchema(etal.LabelsSchema):
         type: the fully-qualified name of the Attribute class
         exclusive: whether at most one attribute with this name may appear in
             an AttributeContainer
+        default: an optional default value for the attribute
     '''
 
-    def __init__(self, name, exclusive=False):
+    def __init__(self, name, exclusive=False, default=None):
```
benjaminpkane: So can an `AttributeSchema` have a default value that is not in its schema?

brimoor: Good point- I added validation for that

PR #458, opened 2020-03-23
Adding support for building BoundingBoxes and RelativePoints from absolute coordinates

brimoor:
Tested by running inference with a TF-Models model that uses the new functionality:

```py
import eta.core.image as etai
import eta.core.learning as etal

# Load model
detector = etal.load_default_deployment_model("ssd-mobilenet-v1-coco")

img = etai.read(...)

with detector:
    objects = detector.detect(img)
```


brimoor:
`coords_in` is the inverse from `from_coords`:
- `coords_in` = relative -> absolute
- `from_coords` = absolute -> relative

PR #461, opened 2020-04-08
Auto-docs for ETA via Sphinx

brimoor:
Enabling auto-generated documentation for ETA via Sphinx. Could use a bit of optimization, but it works!

In doing so, also splitting dev requirements like Sphinx into a separate `requirements/dev.txt` file, installed via a new `-d` flag on the install script.

Only 541 warning messages due to malformed docs... any volunteers to fix? :)

To generate the docs yourself:

```bash
# Install developer dependencies
bash install.bash -d

# Generate the docs
bash sphinx/generate_docs.bash
```

To view the documentation, open `sphinx/build/html/index.html` in your browser.

Example output:
Screen Shot 2020-04-08 at 6 40 54 PM


jasoncorso:
I tried to add inheritance chain graph rendering to the config this morning, but failed.  Didn't have enough time to dig deeper.

PR #466, opened 2020-05-19
Packaging

lethosor:
Currently this just supports a lite installation and an additional `storage` extra for `eta.core.storage`. tensorflow/models is still unhandled by this, but the existing installation process should work for that.


eta/core/utils.py
```diff
@@ -3100,3 +3100,31 @@ def get_terminal_size():
...
+
+class _DeferredImportError(object):
+    def __init__(self, message):
+        self.message = message
+
+    def __getattr__(self, attr):
+        raise ImportError(self.message)
+
+
+def import_defer_error(module_name, message=None):
```
brimoor, line 3113: This is clever!

requirements/common.txt
```diff
@@ -30,5 +30,6 @@ setuptools==45.2.0;python_version>="3"
 simplejson==3.8.1
 six==1.11.0
 sortedcontainers==2.1.0
+Sphinx==2.4.4
```
brimoor, line 33: Sphinx is only for documentation generation, so we should fix whatever underlying problem caused you to upgrade this to a common requirement. It should definitely only be dev installs that need this. `sphinxcontrib-napoleon==0.7` and `sphinx-rtd-theme==0.4.3` are also doc requirements, for example

lethosor: It was this: https://github.com/voxel51/eta/blob/85051ed63df43cafe52e5ea724453f8749d9908d/eta/core/metadata.py#L31-L32
And the CLI imports that module: https://github.com/voxel51/eta/blob/c9a089f87b137504fde9ec0b3f807fc4a0b2c682/eta/core/cli.py#L38
It does appear that only a few commands rely on it, so it could use the same workaround as `eta.core.storage`

brimoor: AHHH, that is required for lite installs actually. The code you linked is used to automatically parse docstrings when generating documentation for ETA modules. So, it 

In fact, judging from the `sphinx.ext.napoleon` import, it seems that `sphinxcontrib-napoleon` needs to be a common requirement as well. The code parses Google style docstrings, which is not something that Sphinx itself knows about

lethosor: `sphinx.ext.napoleon` is installed just as part of `Sphinx` (at least in 3.0). `sphinxcontrib.napoleon` is what `sphinxcontrib-napoleon` installs. I'm not sure if `sphinxcontrib-napoleon` is required under the hood, but nothing else seems to be triggering its installation, and `eta modules --metadata eta/modules/clip_videos.py`

PR #468, opened 2020-05-22
Utility function to take a window snap-shot programmatically.

jasoncorso:
`eta.core.utils.save_window_snapshot` is now available to programmatically save a window to an image file.  Only tested on Linux but works like a charm there.

### DIFF
--- a/eta/core/utils.py
+++ b/eta/core/utils.py
@@ -3412,13 +3412,17 @@ def get_terminal_size():
         return w, h
 
 
-def save_window_snapshot(window_name=None, file_path=None):
+def save_window_snapshot(window_name, file_path):
     """Take a screenshot of a window of a given name and save it to an image on
     disk.
+
+    Args:
+        window_name: the name of the window
+        file_path: the path to write the image on disk
     """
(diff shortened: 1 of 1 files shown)


eta/core/utils.py
```diff
@@ -3410,3 +3410,15 @@ def get_terminal_size():
             ),
         )
         return w, h
+
+
+def save_window_snapshot(window_name=None, file_path=None):
```
brimoor, line 3415: Why add `=None` if they aren't actually optional?

eta/core/utils.py
```diff
@@ -3410,3 +3410,15 @@ def get_terminal_size():
...
+    """Take a screenshot of a window of a given name and save it to an image on
+    disk.
```
brimoor, line 3417: For Sphinx friendliness:

```py
def save_window_snapshot(window_name, file_path):
    """Takes a screenshot of the given window and saves it as an image on disk.

    Args:
        window_name: the name of the window
        file_path: the path to write the image on disk
    """
```

PR #470, opened 2020-06-17
Windows fixes

lethosor:


eta/core/utils.py
```diff
@@ -1195,7 +1196,7 @@ def close(self):
         self._is_running = False
         self._is_finalized = True
 
-        if self.has_dynamic_width:
+        if self.has_dynamic_width and hasattr(signal, "SIGWINCH"):
```
brimoor, line 1199: just make `has_dynamic_width` False in this case?

lethosor: I suppose that does make more sense, as long as SIGWINCH doesn't get used anywhere else

lethosor: Actually, the other case does need has_dynamic_width to be True to adjust the size when the progress bar is created (and that part works on Windows, at least): https://github.com/voxel51/eta/blob/04711ed44f08efd49cf4434674373b1ee2238ca4/eta/core/utils.py#L1022-L1025
I think the current approach will make it easier to add resizing support on Windows later, but you've dealt with the progress bar stuff a lot, so it's up to you.

brimoor: No preference

PR #480, opened 2020-09-28
Adds support for storing and rendering keypoints and polylines

brimoor:
See https://github.com/voxel51/fiftyone/pull/562 for reasonably thorough testing of the functionality introduced in this PR.

Standalone example usage:

```py
import random

import eta.core.image as etai
import eta.core.keypoints as etak
import eta.core.polylines as etap


IMAGE_PATH = "/path/to/test-white.png"
ANNO_PATH = "/path/to/test-anno.png"

img = etai.read(IMAGE_PATH)

image_labels = etai.ImageLabels()

keypoints = etak.Keypoints(
    label="random",
    points=[(random.random(), random.random()) for _ in range(10)]
)
image_labels.add_keypoints(keypoints)

polyline = etap.Polyline(
    points=[(0.3, 0.3), (0.7, 0.3), (0.7, 0.7)], closed=False, filled=False
)
image_labels.add_polyline(polyline)

polyline = etap.Polyline(
    label="triangle",
    points=[(0.4, 0.4), (0.6, 0.4), (0.6, 0.6)],
    closed=True,
    filled=True
)
image_labels.add_polyline(polyline)

print(image_labels)
print(image_labels.get_active_schema())

etai.write(img_anno, ANNO_PATH)
```

test-anno


eta/core/annotations.py
```diff
@@ -867,26 +1296,27 @@ def _draw_object(img, obj, annotation_config, color=None, pre_attrs=None):
...
-    )
+    ) and not annotation_config.hide_all_confidences
     show_index = annotation_config.show_object_indices
+    show_name_only_titles = annotation_config.show_name_only_titles
     occluded_attr = annotation_config.occluded_object_attr
     hide_occluded = annotation_config.hide_occluded_objects
-    labels_whitelist = annotation_config.object_labels_whitelist
-    labels_blacklist = annotation_config.object_labels_blacklist
+    labels_whitelist = annotation_config.labels_whitelist
+    labels_blacklist = annotation_config.labels_blacklist
```
lethosor, line 1319: A versioning question: Since the field name change breaks compatibility (as far as I can tell), this would technically require a bump to `0.2.0`. Is that ok with you, or would you rather keep compatibility?
If this change doesn't break compatibility with FiftyOne/other library users, or if it's a small-enough change that it would go unnoticed, I would be fine sticking with `0.1.x`.

brimoor: I'm not too worried about backwards compatibility on this change, since it's unlikely that anyone is using this particular. Also, it wouldn't actually cause an error, only a silent ignore of a parameter:

```py
from eta.core.annotations import AnnotationConfig  # or fiftyone.utils.annotations

AnnotationConfig({"foo": "bar"})  # non-existent parameter gets silently ignored
```

brimoor: So I say stick with `0.1.X` here

lethosor: Good to know, that makes sense

lethosor (approved):

PR #481, opened 2020-09-30
Progress bar fixes

lethosor:
- Changed quiet progress bars to skip the window size check entirely
- Added error suppression for window resize signal handlers, since they're installed on a best-effort basis already

Needed by https://github.com/voxel51/fiftyone/pull/559 (see https://github.com/voxel51/fiftyone/pull/559#discussion_r497598500 for details)


brimoor (approved):
LGTM

PR #485, opened 2020-10-08
Fix ProgressBar errors when stdout isn't UTF-8

lethosor:
Notably, occurs in some Windows environments, but can be reproduced on Linux by setting the `LC_CTYPE=C` environment variable.

See https://github.com/voxel51/fiftyone/runs/1227722383?check_suite_focus=true#step:16:111 for an example error from before this change


brimoor (approved):
LGTM

PR #487, opened 2020-10-14
Adding support for storing multiple shapes in a Polyline instance

brimoor:
Data formats like COCO and BDD allow for the possibility of a single "object" being composed of multiple disjoint polylines/polygons. This PR updates the data model to allow for this. IE `Polyline.points` is now a list of lists of `(x, y)` points, where each outer list defines a separate shape.

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -975,7 +975,8 @@ def _draw_polyline(img, polyline, annotation_config):
 
     # Render coordinates for image
     # Note: OpenCV expects numpy arrays
-    points = np.array(polyline.coords_in(img=img), dtype=np.int32)
+    points = polyline.coords_in(img=img)
+    points = [np.array(shape, dtype=np.int32) for shape in points]
 
     #
     # Draw polyline
@@ -1000,7 +1001,7 @@ def _draw_polyline(img, polyline, annotation_config):
     # Draw title string
     #
 
-    tcx, tcy = tuple(np.mean(points, axis=(0, 1)))
+    tcx, tcy = np.mean([np.mean(shape, axis=0) for shape in points], axis=0)
     ttlx, ttly, tw, th = _get_panel_coords(
         [title_str], annotation_config, center_coords=(tcx, tcy)
     )
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -954,7 +954,7 @@ def _draw_polyline(img, polyline, annotation_config):
         if polyline.label in labels_blacklist:
             return_now = True
 
-    if not polyline.points:
+    if not polyline.has_vertices:
         return_now = True
 
     if return_now:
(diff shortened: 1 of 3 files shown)


lethosor:
There doesn't appear to be error checking for polylines using the old format. Should there be? Do they work? I assume that draw_annotations and other backend utilities would only handle the new format, but player51 could handle both fairly easily.

```python
fo.Polyline(points=[(0.1, 0.1), (0.3, 0.1), (0.3, 0.3), (0.1, 0.3)])
fo.Polyline(points=[[(0.1, 0.1), (0.3, 0.1), (0.3, 0.3), (0.1, 0.3)]])
```

brimoor:
Polylines in the old format (which has only existed for a week) won't work. I'd rather not support two formats simultaneously in the data model itself. The preference would be to instead add a factory method to create a `Polyline` from a single shape. But, it's easy enough to do `Polyline(points=[single_shape])` at creation time, so I don't think its warranted.

I'd prefer to think of this as a one-time backwards incompatible change

lethosor:
Okay, my question is whether my first example should throw an error, because it currently doesn't. (This polyline even displays successfully in the app, which hasn't been updated to use the new format yet.)
```
fo.Polyline(points=[(0.1, 0.1), (0.3, 0.1), (0.3, 0.3), (0.1, 0.3)])
```

lethosor:
Sorry, my comments were meant to be on https://github.com/voxel51/fiftyone/pull/601. Will continue there.

lethosor:
Another thought: the `closed` and `filled` attributes (and others) are intentionally tied to the entire Polyline, not individual shapes, correct?

brimoor:
yes that was intentional because one can always use `PolylineContainer()` to store multiple polylines with heterogenous rules for `closed/filled`.

My rule for shapes within a single `Polyline` is that they should be "semantically related". If one is filled and the other isn't, I'd argue that they are not "semantically related" and should be stored as separate `Polyline`s.

The typical use case for multiple shapes within a `Polyline` would be an object that is occluded and requires more than one disconnected-but-homogeneous region to cover it.

Comparing CVAT/COCO/BDD formats, they all have incompatible data models for polylines. CVAT strictly allows one shape per instance, which is definitely bad (in general). COCO allows multiple shapes but they're all implicitly closed and filled. BDD allows multiple shapes with a single "polyline" that can have different closed/filled and allows for attributes at both the polyline and shape levels.

Anyway we pick, we're breaking invertibility with at least on of those data types. I let it all bounce around in my head for awhile and decided that the chosen implementation here makes the most sense. Allowing for multiple shapes within an "object" was the big thing I wanted to support.

lethosor:
Sounds good - having a single `filled` value certainly makes things easier on player51's end, so I wanted to make sure it would remain that way.

PR #494, opened 2020-12-22
Fixing non-editable install issues

brimoor:
ETA was previously storing important files in the top-level `eta/` directory, which is not included in `pip install voxel51-eta` installs. Fix PR migrates all of these files to `eta/eta/` so that they are accessible for global installs.


benjaminpkane (approved):
Build looks good!

brimoor:
@benjaminpkane feel free to bump the version number and merge + build this when you're ready. I'll let you handle that

benjaminpkane:
> @benjaminpkane feel free to bump the version number and merge + build this when you're ready. I'll let you handle that

Will do

PR #496, opened 2021-01-06
TF2 object detection model zoo support

ehofesmann:
Added support for detection models from the TF2 detection model zoo: https://github.com/tensorflow/models/blob/master/research/object_detection/g3doc/tf2_detection_zoo.md

Only CenterNet has been implemented with this but the saved model format should match that of the RCNN and SSD [meta architectures](https://github.com/tensorflow/models/tree/master/research/object_detection/meta_architectures) as well.

This does not yet support features or probabilities, but those should be able to be added in a future PR.

**Example usage**

```
pip install --upgrade tensorflow
eta install models
```

```py
import eta.core.image as etai
import eta.core.learning as etal

IMAGE_PATH = "/path/to/image.png"
ANNO_PATH = "/path/for/image-anno.png"

model = etal.load_default_deployment_model("centernet-hg104-512-coco")  # new model
#model = etal.load_default_deployment_model("ssd-mobilenet-v1-coco")  # previously required tensorflow<2

model.config.confidence_thresh = 0.3

with model:
    img = etai.read(IMAGE_PATH)
    objects = model.detect(img)
    image_labels = etai.ImageLabels(objects=objects)

etai.write(img_anno, ANNO_PATH)
```


brimoor (changes requested):
Did you mean to add the CenterNet model to `eta/models/manifest.json`?

eta/detectors/tfmodels_detectors.py
```diff
@@ -385,6 +382,157 @@ def _evaluate(self, imgs, ops):
...
+            detections
+    """
+
+    def __init__(self, d):
+        d = self.init(d)
+
+        self.labels_path = etau.fill_config_patterns(
+            self.parse_string(d, "labels_path")
+        )
+        self.pipeline_path = etau.fill_config_patterns(
```
brimoor, line 408: `pipeline_path` appears to be unused here

eta/detectors/tfmodels_detectors.py
```diff
@@ -385,6 +382,157 @@ def _evaluate(self, imgs, ops):
...
+    """TF2ModelsDetector configuration settings.
+
+    Note that `labels_path` is passed through
+    `eta.core.utils.fill_config_patterns` at load time, so it can contain
+    patterns to be resolved.
+
+    Attributes:
+        model_name: the name of the published model to load. If this value is
+            provided, `model_path` does not need to be
+        model_path: the path to a frozen inference graph to load. If this value
```
brimoor, line 395: I think this now must be an `archive` of a `SavedModel` directory or something like that

brimoor:
See also this https://github.com/voxel51/fiftyone/pull/740#issuecomment-756329511 for question regarding using TF slim models after this update directly in ETA.

eta/detectors/tfmodels_detectors.py
ehofesmann: Thought I deleted that, removed from the FiftyOne `manifest.json` as well.

eta/detectors/tfmodels_detectors.py
ehofesmann: Right. On a different note, is there something that I need to do with `model_name` or is does it automatically populate  `model_path` if it is provided instead?

ehofesmann:
> Did you mean to add the CenterNet model to `eta/models/manifest.json`?

I didn't intend to but I can if you think it should be in there.

eta/detectors/tfmodels_detectors.py
brimoor: `model_name` and `model_path` are defined by the `eta.core.learning.HasPublishedModel` interface. When adding models to `eta/models/manifest.json`, you should populate `model_name` and not `model_path` in the default deployment config for the model. When someone loads the model via `etal.load_default_deployment_model()`, then the model's weights will be downloaded and `model_path` will be populated at runtime.

You'll notice that in FiftyOne, and model `manifest.json` files don't include `model_name` or `model_path` in the default deployment config. The appropriate value (for `model_path` in this case, not `model_name`, since we want to let FiftyOne control where the models are downloaded, not ETA) is inserted by `load_zoo_model()` when someone loads a model.

brimoor:
> > Did you mean to add the CenterNet model to `eta/models/manifest.json`?
> 
> I didn't intend to but I can if you think it should be in there.

It'd be nice to have something in there so the actual TF2 functionality can be tested in this PR!

ehofesmann:
> > > Did you mean to add the CenterNet model to `eta/models/manifest.json`?
> > 
> > 
> > I didn't intend to but I can if you think it should be in there.
> 
> It'd be nice to have something in there so the actual TF2 functionality can be tested in this PR!

Tested this and it works:

```
import eta.core.image as etai
import eta.core.learning as etal

# must run `eta install models` first
model = etal.load_default_deployment_model("centernet-hg105-512-coco-tf2", error_level=1)

with model:
    img = etai.read("/path/to/image.png")
    print(model.predict(img))
```

eta/detectors/tfmodels_detectors.py
```diff
@@ -435,15 +437,21 @@ def __init__(self, config):
...
         # Load labels
         self._category_index, self._class_labels = _parse_labels_map(
             self.config.labels_path
         )
         self._num_classes = len(self._class_labels)
 
+        self._detect_fn = None
+
+    def __enter__(self):
+        self._detect_fn = _load_tf2_detection_model(self._model_dir)
```
ehofesmann, line 449: Just curious, why would you rather load the model here than in the constructor?

eta/models/manifest.json
```diff
@@ -917,15 +917,47 @@
...
+                }
+            },
+            "requirements": {
+                "cpu": {
+                    "support": true,
+                    "packages": ["tensorflow>=2"]
+                },
+                "gpu": {
+                    "support": true,
+                    "packages": ["tensorflow-gpu>=2"]
```
brimoor, line 956: Didn't they get rid of `tensorflow-gpu` starting in TF2? I think you may be able to do this instead:
```
            "requirements": {
                "packages": "tensorflow>=2",
                "cpu": {
                    "support": true,
                },
                "gpu": {
                    "support": true
                }
            }
```

ehofesmann: Well would you look at that, this is the first I've heard of it: https://www.tensorflow.org/install/gpu#pip_package

eta/detectors/tfmodels_detectors.py
brimoor: All the other models do it this way too. The idea is to do all the expensive loading in `__enter__` and then cleanup in `__exit__` so that someone could run lots of models in series without graph conflicts (in TF1) or running out of memory (in general).

IE I'd like for someone to be able to do this:

```py
for model_name in etam.list_models():
    model = etal.load_default_deployment_model(model_name)
    with model:
        model.detect(img)
```

On that note, is it necessary and/or possible to cleanup a model that's been loaded with `tf.saved_model.load()`?

brimoor: I think the answer to my question is _supposed_ to be "no: tensorflow 2 will garbage collect automatically". but stack overflow indicates this may not always work sufficiently well. 🤷‍♂️

eta/models/manifest.json
```diff
@@ -373,11 +373,11 @@
             "requirements": {
                 "cpu": {
                     "support": true,
-                    "packages": ["tensorflow<2"]
+                    "packages": ["tensorflow"]
                 },
                 "gpu": {
                     "support": true,
-                    "packages": ["tensorflow-gpu<2"]
+                    "packages": ["tensorflow-gpu"]
```
brimoor, line 380: Does `tensorflow-gpu` work for version `>=2`? The official docs say `tensorflow>=2` will include GPU support, but the problem is that these requirements should technically read `tensorflow-gpu<2 OR tensorflow>=2`, but I'm not sure if `pip` allows package specifications with `OR` in them, and I don't really want to write our own code for that...

ehofesmann: You can `pip install tensorflow-gpu==2` so it should work, no?

brimoor: But users are likely to have already installed TF 2.X via `pip install --upgrade tensorflow` like the official docs recommend. As is, they'd get a "`tensorflow-gpu` not installed" error here. Although they could set `error_level` to ignore it, I'd rather our code not force people to use workarounds when following the "official" way to use other packages.

I'm going to tweak our package verification code to allow the `<requirement1>|<requirement2>` syntax

brimoor (approved):
Okay I added support for allowing either `tensorflow-gpu<2` or `tensorflow>=2`, for example.

If such a requirement is violated, the user gets a pretty informative message. For example:

```py
import eta.core.utils as etau
etau.ensure_package("tensorflow>=2|tensorflow-gpu<2")
```
```
ImportError: The requested operation requires that one of the following is installed on your machine:
-   tensorflow>=2
-   tensorflow-gpu<2

but found:
-   tensorflow==1.15.5
```

I've tested a representative example of each model type under both `tensorflow<2` and `tensorflow>=2` and confirmed that this PR is working as expected.

PR #498, opened 2021-01-08
Bump to v0.3.0

benjaminpkane:
I'd like to start following the practice of bumping the minor (to `3` in this case) for better semantic versioning and requirement handling in FiftyOne.


brimoor (approved):
LGTM

PR #507, opened 2021-03-23
Editable install for development installs only

benjaminpkane:
The motivation here is allow installs from source using the install script in Google's Colab. Installing in Colab with an `-e` flag does not work.

So, in short, this change allows one to install ETA in a Colab notebook with `bash install.bash` but not `bash install.bash -d`.


brimoor (approved):
LGTM

PR #508, opened 2021-03-24
Indicate whether requirement installs were successful

brimoor:
In situations when `error_level > 0`, it can be useful to get a return flag indicating whether a particular requirement is installed.

For example, third-party code may want to set `error_level=1` and then issue an additional message suggesting a workaround if a package is not installed.

```py
import eta.core.utils as etau

print(etau.install_package("tensorflow"))  # True
print(etau.install_package("tensorflowz", error_level=1))  # False

print(etau.ensure_package("tensorflow"))  # True
print(etau.ensure_package("tensorflowz", error_level=1))  # False

print(etau.ensure_import("tensorflow"))  # True
print(etau.ensure_import("tensorflowz", error_level=1))  # False

print(etau.ensure_cuda_version(">51"), error_level=1)  # False
print(etau.ensure_cudnn_version(">51"), error_level=1)  # False
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ complete information.
 ## Quickstart
 
 Get your feet wet with ETA by running some of examples in the
-[examples folder](https://github.com/voxel51/eta/tree/develop/examples).
+[examples folder](https://github.com/voxel51/eta/tree/develop/eta/examples).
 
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #510, opened 2021-03-30
Add tags to label objects

benjaminpkane:
Adds tags to FrameLabels (segmentations), Attributes, DetectedObjects, Polylines, and Keypoints.


brimoor (approved):
LGTM

PR #517, opened 2021-06-23
Hotfixes for FiftyOne v0.10.0

brimoor:
Fixes a bug that was causing calls like `fiftyone.core.utils.ensure_import("pycocotools")` to raise uncaught exceptions.

Also adds missing requirements files used by `install.bash` for source installs that were intended to be included in #515.


benjaminpkane (approved):
LGTM

PR #518, opened 2021-07-22
Fix bug when computing mask from polyline

ehofesmann:
Numpy arrays are `(height, width)` but the mask was being initialized with the `(width, height)` mask size.


brimoor (approved):
LGTM

PR #520, opened 2021-08-06
Potential solution for issue 1163 in fiftyone

ErfanTagh:
Added proxy settings for users connecting from countries were they need a proxy to access Google Drive

### DIFF
--- a/eta/core/web.py
+++ b/eta/core/web.py
@@ -23,6 +23,9 @@ import logging
 import re
 import requests
 
+import urllib.request
+
+
 import eta.constants as etac
 import eta.core.utils as etau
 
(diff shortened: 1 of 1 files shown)


eta/core/web.py
```diff
@@ -150,12 +153,16 @@ def write(self, path, url, params=None):
             self._do_download(r, f)
 
     def _get_streaming_response(self, url, headers=None, params=None):
+        #get the defualt system proxy settings by default
+        proxies = urllib.request.getproxies()
```
brimoor, line 157: Would it make more sense to register the proxies only once when `sess` is created in the class constructor?

brimoor (commented):
Thanks for taking a look into this @ErfanTagh! I left a comment for you and will also see if I can get the OP from https://github.com/voxel51/fiftyone/issues/1163 to test this for us 😄

brimoor:
Not needed- see https://github.com/voxel51/fiftyone/issues/1163#issuecomment-894933905.

PR #522, opened 2021-09-08
Adding support for drawing annotated clips

brimoor:
Adds an optional `support` parameter to `annotate_video()` that can be used to render an annotated version of a specific clip of a video.

This is used by https://github.com/voxel51/fiftyone/pull/1228.


benjaminpkane (approved):
LGTM

PR #523, opened 2021-09-08
Render polylines of different sizes

ehofesmann:
This PR fixes a bug where `eta.core.image.render_bounding_box_and_mask()` only allows for polylines where every shape that it contains has the same number of points. It now mimics what is done in `fiftyone.core.labels._render_polyline()`.


brimoor (approved):
LGTM

PR #527, opened 2021-10-23
Allow directory to exist when running ensure_dir()

brimoor:
Tweaks the implementation of `ensure_dir()` so that it is **always** okay if a directory already exists. Previously I had errors in multiprocess workflows where one worker managed to create the directory between the execution of the `if` and `os.makedirs()` statements of another worker.


benjaminpkane (approved):
LGTM

PR #532, opened 2021-11-18
Config and AWS profile enhancements

brimoor:
Change log
- Adds methods for parsing path config variables that automatically expand `~` and relative paths
- Adds support for custom AWS profiles in both config (previously supported) and credential (now supported formats)


benjaminpkane (approved):
LGTM

PR #533, opened 2021-11-23
Adding a MinIOStorageClient

brimoor:
Adds a `MinIOStorageClient` that supports working with [MinIO](https://min.io) storage deployments.

MinIO is fully boto-compliant, so the primary work was providing a mechanism to configure MinIO deployment details via environment variables and credential files.


benjaminpkane (approved):
LGTM

PR #535, opened 2021-12-03
Adding is_folder()

brimoor:
Adds an `StorageClient.is_folder()` method for efficiently determining whether a given remote "folder" exists (ie is non-empty).


benjaminpkane (approved):
LGTM

PR #544, opened 2022-01-19
Adding retry policy to HTTPStorageClient

brimoor:
Adds a retry policy to `HTTPStorageClient` that behaves similarly to the default retry policies of `S3StorageClient` and `GCSStorageClient`.


benjaminpkane (approved):
LGTM

PR #545, opened 2022-01-20
Adding tags to VideoEvents

brimoor:
Better consistency with `fiftyone.core.labels.TemporalDetection` type.


benjaminpkane (approved):
LGTM

PR #547, opened 2022-01-26
Adding support for pulling AWS credentials from an IAM role

brimoor:
In certain workflows, AWS users may want to interact with S3 using credentials inherited from an IAM role as exposed at runtime by the following environment variables:

```
AWS_ROLE_ARN
AWS_WEB_IDENTITY_TOKEN_FILE
```

Unfortunately, credentials generated via this method are strictly temporary. Fortunately, there is a `botocore.credentials.RefreshableCredentials` class that provides a convenient way to automatically regenerate credentials when needed.

Helpful resources
- https://dev.to/li_chastina/auto-refresh-aws-tokens-using-iam-role-and-boto3-2cjf
- https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iam.html#IAM.Client.get_role
- https://stackoverflow.com/q/60747733

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -681,6 +681,7 @@ class _BotoStorageClient(StorageClient, CanSyncDirectories):
         self.alias = alias
         self.endpoint_url = endpoint_url
 
+        self._prefixes = tuple(prefixes)
         self._role_arn = None
         self._web_identity_token = None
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM

PR #550, opened 2022-02-07
read json file with utf-8

MagicCodess:
Different language platform especially in windows may cause error here, read json file with utf-8 decoder might work better.

### DIFF
--- a/eta/core/serial.py
+++ b/eta/core/serial.py
@@ -103,7 +103,7 @@ def read_json(path):
         ValueError: if the JSON file was invalid
     """
     try:
-        with open(path, "rt") as f:
+        with open(path, "rt", encoding='utf-8') as f:
             return json.load(f)
     except ValueError:
         raise ValueError("Unable to parse JSON file '%s'" % path)


brimoor (commented):
Hi @MagicCodess 👋 I think a more robust solution to this problem for you is to instead set your [locale preferred encoding](https://stackoverflow.com/q/31469707) to UTF-8.

Making this change may resolve a particular error for you, but it won't necessarily resolve all such errors when using this library or upstream libraries like [FiftyOne](https://github.com/voxel51/fiftyone). Also, this library doesn't necessarily have a UTF-8 only assumption, so I'd need to consider this a bit more globally to see what the broader implications might be to make a proper promise on string encoding support.

MagicCodess:
hi @brimoor :), I think the graceful way for this is to detect the json file encoding, in this way users don't need to change their locale encoding which may cause some errors in their other apps. I have committed some new codes which using chardet to detect file encoding, thanks for your reviewing.

brimoor (commented):
Hmm while this is fancy, in the worst case it could make loading JSON 2x slower...

I think I'm inclined to go with your UTF-8-only assumption instead for simplicity. But, again, we'll need to check for more places where this assumption may need to be implemented. For example `write_json()` in this library. But also in [FiftyOne](https://github.com/voxel51/fiftyone) which is the primary dependency of this library.

PR #551, opened 2022-02-09
Always reload web identity token file

brimoor:
When working with `S3StorageClient` using the `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE`, it turns out that the web identity token file is updated periodically.

This PR re-reads the web identity token each time credentials are refreshed to ensure that expiration never occurs.

Resources
- https://aws.github.io/aws-eks-best-practices/security/docs/iam/#iam-roles-for-service-accounts-irsa
- https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -684,7 +684,7 @@ class _BotoStorageClient(StorageClient, CanSyncDirectories):
 
         self._prefixes = tuple(prefixes)
         self._role_arn = None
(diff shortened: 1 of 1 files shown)


benjaminpkane (approved):
LGTM

PR #552, opened 2022-02-09
Remote storage client path tweaks

brimoor:
Changelog
- Always use `/` rather than `os.path.sep` when dealing with remote paths, since these are URLs or S3/GCS/MinIO paths, which always use `/`
- Retain the user's prefix in `_BotoStorageClient.list_files_in_folder()`. Previously listing a MinIO folder using the `http://` syntax would return `alias://` paths if the user configured an alias; however, it is more natural to use the same prefix that the user actually provided


benjaminpkane (approved):
LGTM

PR #562, opened 2022-05-25
adding explicit start_number to ffmpeg

neuralsyn:
When ffmpeg is searching for files it looks for a sequence pattern containing `"%d"` or `"%0Nd"`, where the first filename of the file list specified by the pattern must contain a number inclusively contained between `start_number` and `start_number+start_number_range-1`.  By [default](http://ffmpeg.org/ffmpeg-all.html#image2-1) `start_number=0` and `start_number_range=5`.  If a file sequence begins at or after `start_number_range=5` an exception is thrown:
```
Could find no file with path '/fake/path/%06d.jpg' and index in the range 0-4
```

This change automatically determines the `start_number` to use based on the ffmpeg `inpath` parameter.  If no sequence is found then it will default to `0`, otherwise it will use the result of sequence pattern parsing.

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -1957,14 +1957,12 @@ class VideoSetLabels(etal.LabelsSet):
         self.sort_by("filename", reverse=reverse)
 
     def clear_video_attributes(self):
-        """Removes all video-level attributes from all VideoLabels in the set.
-        """
+        """Removes all video-level attributes from all VideoLabels in the set."""
         for video_labels in self:
             video_labels.clear_video_attributes()
 
     def clear_frame_attributes(self):
-        """Removes all frame-level attributes from all VideoLabels in the set.
-        """
+        """Removes all frame-level attributes from all VideoLabels in the set."""
         for video_labels in self:
             video_labels.clear_frame_attributes()
 
@@ -4138,6 +4136,13 @@ class FFmpeg(object):
             in_opts = self.DEFAULT_IN_OPTS
         else:
             in_opts = self._in_opts
+
+        # Automatically determine the starting number of the inpath for cases
+        # where it's a sequence pattern (e.g. %06d.jpg).  The default behavior
+        # from ffmpeg is to start at 0 and look in the range [0,4].  If the first
+        # matched pattern begins above 4 we want to explicitly set that.
+        start_number = next(iter(etau.parse_pattern(inpath)), None) or 0
+        in_opts.extend(("-start_number", str(start_number)))
 
         # Output options
         if self._out_opts is None:


neuralsyn:
Additional testing revealed that the parameter is unintentionally added multiple times during calls to `to_frames`.  Working on a fix now.

neuralsyn:
Added handling for duplicate options and inpath sequence pattern checking.  Ready for someone to take a look now.

eta/core/video.py
```diff
@@ -4139,6 +4137,16 @@ def run(self, inpath, outpath, verbose=False):
...
 
+        # Automatically determine the starting number of the inpath for cases
+        # where it's a sequence pattern (e.g. %06d.jpg).  The default behavior
+        # from ffmpeg is to start at 0 and look in the range [0,4].  If the first
+        # matched pattern begins above 4 we want to explicitly set that.
+        if "-start_number" not in in_opts and is_supported_image_sequence(
+            inpath
+        ):
+            start_number = next(iter(etau.parse_pattern(inpath)), 0)
+            in_opts.extend(["-start_number", str(start_number)])
```
brimoor, line 4148: The design of the `FFmpeg` class is that you provide `in_opts` and `out_opts` at class instantiation and can then call `run()` one or more times to use those parameters on multiple videos.

Given that, I think you should include `in_opts = in_opts.copy()` here if you are going to modify the options based on a particular call to `run()`. Maybe the user calls `run()` again with a non-image-sequence input, for example. They wouldn't want `-start_number` in that call.

Alternatively, you could push the responsibility of knowing that `-start_number` needs to be included to the caller of `FFmpeg(in_opts=[...])`.

The `FFmpeg` class here was intended to be a very low-level wrapper around the `ffmpeg` binary, very little magic happening without the caller explicitly requesting it. Although in this particular case, I'm onboard with a little magic to make `/path/to/frames/%06d.jpg` work better, if you want to put it here.

eta/core/video.py
```diff
@@ -4139,6 +4137,16 @@ def run(self, inpath, outpath, verbose=False):
...
             in_opts = self._in_opts
 
+        # Automatically determine the starting number of the inpath for cases
+        # where it's a sequence pattern (e.g. %06d.jpg).  The default behavior
+        # from ffmpeg is to start at 0 and look in the range [0,4].  If the first
+        # matched pattern begins above 4 we want to explicitly set that.
+        if "-start_number" not in in_opts and is_supported_image_sequence(
+            inpath
+        ):
+            start_number = next(iter(etau.parse_pattern(inpath)), 0)
```
brimoor, line 4147: Since `FFmpeg` is supposed to be a very low-level wrapper around `ffmpeg`, consider putting the `start_number` computation inside a `try-except` so that, if we fail to guess a start number, the command runs nonetheless, in case `ffmpeg` is smarter? (perhaps with a warning logged).

eta/core/video.py
neuralsyn: The condition where the `is_supported_image_sequence` method is checked accounts for the scenario where an image sequence is *not* provided.  As-is the only time this code runs is explicitly when an image sequence has been provided and the user did not themselves specify a `-start_number`.  The "magic" is that in such cases we automatically determine a starting point, however this point may in fact still be the ffmpeg default of 0 (i.e. the default specified in `next`) which won't cause any unexpected side-effects given that it's the built-in default.  In all other scenarios this behavior will be skipped entirely.

eta/core/video.py
brimoor: Ah missed the `0`. My bad

eta/core/video.py
brimoor: I get that, I'm just pointing out that the current implementation modifies either the `FFmpeg.DEFAULT_IN_OPTS` or `FFmpeg._in_opts` list in-place, doesn't it?

```py
ffmpeg = FFmpeg()

# Oops, permanently modifies `FFmpeg.DEFAULT_IN_OPTS`
ffmpeg.run("/path/to/frames/%06d.jpg", "/path/for/video1.mp4")

# Doesn't this wrongly include `start_number` in `in_opts`?
ffmpeg.run("/path/to/video.mov", "/path/for/video2.mp4")
```

neuralsyn: Ah I see what you're driving at.  I'll push up a fix that calls the `copy()` method that resolves this issue

PR #563, opened 2022-05-26
adding retry to gcs client upload and download methods

Anddraca:
Small update to make sure that the GCS storage client upload and download methods are using the Retry object defined on the class.

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -1882,7 +1882,9 @@ class GoogleCloudStorageClient(
         if metadata:
             blob.metadata = metadata
 
-        blob.upload_from_filename(local_path, content_type=content_type)
+        blob.upload_from_filename(
+            local_path, content_type=content_type, retry=self._retry
+        )
 
     def upload_bytes(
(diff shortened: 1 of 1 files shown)


brimoor (changes requested):
I see some other blob methods used like `blob.delete()`, `blob.patch()`, and `blob.reload()`. Do they need retries too?

brimoor:
nit: PR conventions:
- Add @voxel51/developers (or specific people, if appropriate) as reviewers
- Add yourself (or whoever should merge the PR once it is approved) as assignee
- Add appropriate labels to track type of work (for metrics, to guide reviewers on whether they are appropriate reviewers, etc)

Anddraca:
Thanks for the points on setting up the PR -- I had assigned myself on a different issue but missed this one :grimacing: 

As for the other methods: those do all take the retry object. I'll add those on if that's the desired behavior

brimoor (approved):
LGTM. Thanks for catching the incomplete use of retries here!

Before merging, if there are any other GCS-related calls here that could fail due to errors we're deeming as retryable, please add retries for those too 🤗

PR #565, opened 2022-06-28
Fix start number bug from #562

brimoor:
Looks like https://github.com/voxel51/eta/pull/562 wasn't tested before releasing, because there was a bug (`extend()` returns `None`, not the list).

Example usage:

```py
import eta.core.video as etav

inpath = "/path/to/video.mp4"
frames_patt = "/tmp/frames/%06d.jpg"
outpath = "tmp/video.mp4"

ffmpeg = etav.FFmpeg()
ffmpeg.run(inpath, frames_patt)
ffmpeg.run(frames_patt, outpath)  # fails on `develop`, works on this branch
```

### DIFF
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -4145,7 +4145,9 @@ class FFmpeg(object):
             "-start_number" not in in_opts
         ):
             start_number = next(iter(etau.parse_pattern(inpath)), 0)
-            in_opts = in_opts.copy() + ["-start_number", str(start_number)]
+
+            # Note that `in_opts` modification must be a per-call change
+            in_opts = in_opts + ["-start_number", str(start_number)]
 
         # Output options
         if self._out_opts is None:
--- a/eta/core/video.py
+++ b/eta/core/video.py
@@ -4138,18 +4138,14 @@ class FFmpeg(object):
             in_opts = self._in_opts
 
         # Automatically determine the starting number of the inpath for cases
-        # where it's a sequence pattern (e.g. %06d.jpg).  The default behavior
-        # from ffmpeg is to start at 0 and look in the range [0,4].  If the first
-        # matched pattern begins above 4 we want to explicitly set that.
-        if "-start_number" not in in_opts and is_supported_image_sequence(
-            inpath
+        # where it's a sequence pattern (e.g. %06d.jpg). The default behavior
+        # from ffmpeg is to start at 0 and look in the range [0, 4]. If the
+        # first matched pattern begins above 4 we want to explicitly set that.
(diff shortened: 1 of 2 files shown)


eta/core/video.py
```diff
@@ -4138,18 +4138,14 @@ def run(self, inpath, outpath, verbose=False):
...
+        if is_supported_image_sequence(inpath) and (
+            "-start_number" not in in_opts
         ):
             start_number = next(iter(etau.parse_pattern(inpath)), 0)
-
-            # Ensure in_opts modification is a per-call change
-            in_opts = in_opts.copy().extend(
-                ["-start_number", str(start_number)]
-            )
+            in_opts = in_opts.copy() + ["-start_number", str(start_number)]
```
findtopher, line 4148: nit:

you don't need the `.copy()` here - `+` always creates a new list:
```
In [1]: mylist = ['foo', 'bar']

In [2]: newlist = mylist + ['baz', 'buzz']

In [3]: mylist
Out[3]: ['foo', 'bar']

In [4]: newlist
Out[4]: ['foo', 'bar', 'baz', 'buzz']
```

I would, however, suggest that you consider
`in_opts += ["-start_number", str(start_number)]` as it is significantly more performant.

unless you feel like you need a shallow copy of the list for some reason?

```
In [5]: mylist += ['baz', 'buzz']

In [6]: mylist
Out[6]: ['foo', 'bar', 'baz', 'buzz']
```

for example purposes:
```
In [7]: otherlist = ['foo', 'bar']

In [8]: otherlist = otherlist.copy() + ['baz', 'buzz']

In [9]: otherlist
Out[9]: ['foo', 'bar', 'baz', 'buzz']
```

though you could continue to use `extend` if we got rid of the copy...  that just feels like an odd syntax (very nonpythonic):

```
In [10]: extendlist = ['foo', 'bar']

In [11]: extendlist.extend(['baz', 'buzz'])

In [12]: extendlist
Out[12]: ['foo', 'bar', 'baz', 'buzz']
```

findtopher (approved):
I mean, if it works it works, it just seems like a lot of steps...  Which I'm sure exist for reasons.

eta/core/video.py
findtopher: Yes, I may be bikeshedding list expansion.

brimoor: `FFmpeg` is a bit weird: `in_opts` is shared between calls to `run()` so we need to avoid editing in-place so that each call to `run()` is independent.

Happy to (a) remove `copy()` though and (b) give you a hard time for worrying about list creation right before a subprocess is opened 😉

PR #567, opened 2022-07-12
Disable default logging modifications

brimoor:
Currently, `import eta` triggers `eta.core.logging.basic_setup()` to be called, which modifies the user's root logger. This is bad manners for anyone who is simply using ETA as a dependency and not for it's [module](https://github.com/voxel51/eta/blob/develop/docs/modules_dev_guide.md) or [pipeline](https://github.com/voxel51/eta/blob/develop/docs/pipelines_dev_guide.md) constructs.

This PR removes the default logging configuration so that `import eta` by itself not cause any modifications to the root logger (it is only configured when ETA modules/pipelines are actually run).

### DIFF
--- a/docs/logging_guide.md
+++ b/docs/logging_guide.md
@@ -6,11 +6,7 @@ describes how to use ETA's logging infrastructure.
 
 ## Logging in a module
 
-By convention, ETA names loggers at the module-level, which propagate their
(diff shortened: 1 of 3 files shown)


benjaminpkane (approved):
LGTM

PR #569, opened 2022-09-05
Handling zero height/width when resizing images

brimoor:
`eta.core.image.resize()` was not previously robust to degenerate cases where an input/output width/height was zero. This PR resolves that.

```py
import fiftyone as fo
import numpy as np

detection = fo.Detection(
    label="empty",
    bounding_box=[0, 0, 0, 0],
    mask=np.zeros((0, 0)),
)

# Previously would raise an error; now succeeds
segmentation = detection.to_segmentation(frame_size=(32, 32))
```


benjaminpkane (approved):
LGTM

Issue #579, opened 2023-02-10
Progress bar regression in `v0.8.2`


radao:
The new release seems to have broke the `fo_dataset.add_samples` method. This is with `fiftyone==0.18.0`:
```bash
dataset.add_samples(samples=fo_samples, expand_schema=True, dynamic=False, validate=True)
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/fiftyone/core/dataset.py", line 2209, in add_samples
    for batch in batcher:
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/fiftyone/core/utils.py", line 957, in __iter__
    self._pb.__enter__()
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/eta/core/utils.py", line 1670, in __enter__
    self.start()
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/eta/core/utils.py", line 1829, in start
    self._is_nested = self._cap_obj.__enter__()
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/eta/core/utils.py", line 1391, in __enter__
    self.start()
  File "/root/conda/envs/3.10/lib/python3.10/site-packages/eta/core/utils.py", line 1435, in start
    logger.handlers[idx].stream = self._cache_stdout
AttributeError: can't set attribute 'stream'
```

brimoor:
Hmm, strange I'm not able to reproduce on Python 3.9 or 3.10.

This is the change that was made: https://github.com/voxel51/eta/pull/574/files

The idea is to update the `stream` property of all `logging.StreamHandler` instances in your current logging setup so that the progress bar can be intertwined cleanly with any logging you might be doing within your loop.

Are you able to track down the `type` of whatever `logger.handlers[idx]` is? The code in `_get_stdout_loggers()` should only be grabbing `StreamHandler` objects...

I thought perhaps setting `logging.StreamHandler.stream` was disallowed in Python 3.10 or something, but that doesn't seem to be the case 🤔 

In the meantime, you can bypass this error via:

```py
import fiftyone as fo

fo.config.show_progress_bars = False
```

PR #581, opened 2023-02-11
Directly store stream handlers

brimoor:
Resolves https://github.com/voxel51/eta/issues/579

Well, precisely, I wasn't able to reproduce https://github.com/voxel51/eta/issues/579, but this refactor strictly stores `StreamHandler` instances and uses the `setStream()` method in their [public API](https://docs.python.org/3/library/logging.handlers.html#streamhandler), so IDK what could go wrong 🤷

### DIFF
--- a/eta/core/utils.py
+++ b/eta/core/utils.py
@@ -1368,7 +1368,7 @@ class CaptureStdout(object):
 
     def __init__(self, parent=None):
         """Creates a CaptureStdout instance."""
-        self._stdout_loggers = None
+        self._stdout_handlers = None
         self._orig_stdout = None
         self._cache_stdout = None
         self._stdout_str = None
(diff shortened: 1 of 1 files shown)


eta/core/utils.py
```diff
@@ -1449,31 +1448,26 @@ def stop(self):
...
+        handlers = set()
+        for logger in loggers:
+            for handler in getattr(logger, "handlers", []):
                 if (
                     isinstance(handler, logging.StreamHandler)
-                    and getattr(handler, "stream", None) == sys.stdout
-                )
-            ]
-            if inds:
-                stdout_loggers[logger] = inds
```
kaixi-wang, line 1474: I wasn't able to repro the bug either, but it looks like previously it was setting the logger to an index rather than the handler. It looks correct now.

kaixi-wang (approved):
Lgtm!

PR #583, opened 2023-03-04
Don't raise error when copying same file

brimoor:
Resolves the issue pointed out in https://github.com/voxel51/fiftyone/pull/2670.


swheaton (approved):
lgtm

PR #584, opened 2023-03-08
Fix Kinetics dataset download from YouTube

allenleetc:
Fixes voxel51/fiftyone#2607, fixes voxel51/fiftyone#2473


brimoor (approved):
Ah nice catch. Perhaps we should just use a generic `except` in case there are other types of exceptions that `bad_value % 1` might cause?

PR #585, opened 2023-03-13
[HOLD] Release 0.8.4rc1

brimoor:


setup.py
```diff
@@ -12,7 +12,7 @@
 from wheel.bdist_wheel import bdist_wheel
 
 
-VERSION = "0.8.3"
+VERSION = "0.8.4-rc1"
```
findtopher, line 15: Just as a side note, because I'm pretty sure Version fixes this for you anyway, `0.8.4rc1` (no dash) is what pip will recognize and pull...  

ref https://peps.python.org/pep-0440/#examples-of-compliant-version-schemes

Basically, the package that will be created here will have a different number than what you've put in the file - just in case anybody is ever searching for it.

PR #588, opened 2023-03-20
Azure support

brimoor:


swheaton (commented):
I have not used Azure before but I perused docs and I think this looks good - quite thorough.

I did not create an azure account to try and test this branch though.

Only concern is adding azure requirement to all, even if you are using another provider. It's already bloated because you get both gcloud and AWS. Ideally the extra dependencies would be more fine grained per storage type - e.g., `AzureStorageClient` knows the list of modules to try to import together and then requires `voxel51-eta[azure-storage]` if it fails. I am not 100% sure how to make this flow through to users of this module.
Thoughts? We could make a tech debt ticket to do this in follow up work.

brimoor:
Yeah the `voxel51-eta[storage]` install is definitely bloated, but FiftyOne Teams doesn't use it (it installs requirements itself), so I think we can ignore any bloat in this repo directly.

I think we may want to completely sunset the ETA repo in the near future and just copy any utils we actually use into fiftyone/Teams.

swheaton (approved):

PR #594, opened 2023-04-26
Remove pkg_resources, use importlib metadata instead

### LINKED ISSUE #593: DeprecationWarning: pkg_resources is deprecated as an API
`pkg_resources` is called from `eta.core.utils`, but its use is now deprecated, since setuptools > 67.5

See the warning on most recent file here : https://github.com/pypa/setuptools/blame/main/pkg_resources/__init__.py#L121

See commit introducing the warning here : https://github.com/pypa/setuptools/commit/a1aeda391a0c462ea53627bcdf50dd4c0daadaed

Did not know why I had tons of warnings when running tests for my own library, and found the culprit in this package. I can go back to setuptools 65 the same way pip did : https://github.com/pypa/pip/issues/11975, but might as well prepare for the future.

Recommendation is to use `importlib.resources` instead of `pkg_resources`, see related discussion : https://github.com/simonw/datasette/issues/2057 (nice demonstration of chatgpt pair programming ! :stuck_out_tongue: )

ClementPinard:
Fixes #593

### DIFF
--- a/eta/core/utils.py
+++ b/eta/core/utils.py
@@ -49,7 +49,11 @@ import mimetypes
 import numbers
 import os
 from packaging.requirements import Requirement
-import pkg_resources
+
+try:
+    from importlib import metadata
+except ImportError:
+    import importlib_metadata as metadata
 import py7zr
 import pytz
 import random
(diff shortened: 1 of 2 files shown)


brimoor (approved):
LGTM 🥇

PR #596, opened 2023-05-03
Adding support for models with no managers and requirements.txt files on disk

brimoor:
Changelog
- makes the `manager` property of `eta.core.models.Model` instances optional.
- adds support for loading package requirements for models from a `requirements.txt` file on disk


brimoor:
I've been using this locally for months and it is all pre-FiftyOne code, so I'm just going to merge

PR #599, opened 2023-08-01
Release v0.11.0

brimoor:


findtopher:
In the context of a hosted environment, I'm very uncomfortable with the idea of just being able to pass in requirements and have them automatically installed in the environment...

e.g. `apoclypsebm` seems like it could be a real issue for our hosted environment.

How do we disable this functionality in our hosted environments, or how does a customer limit who can arbitrarily install packages in their environment?  This feels like a great opening for a well-intentioned code-injection attack vector...

benjaminpkane (approved):
LGTM. Will release today

PR #604, opened 2023-12-18
Release/v0.12.1

brimoor:
Adds support for working with class methods in the two methods shown below:

```py
import eta.core.utils as etau
import fiftyone as fo

etau.get_function_name(fo.Dataset.from_dir)
# 'fiftyone.core.dataset.Dataset.from_dir'

etau.get_function('fiftyone.core.dataset.Dataset.from_dir')
# <bound method Dataset.from_dir of <class 'fiftyone.core.dataset.Dataset'>>
```


benjaminpkane (approved):
LGTM

PR #606, opened 2024-01-02
Release v0.12.2

brimoor:
`Pillow==10` removed a `Font.getsize()` method that we used in one module. This PR ensures that we remain compatible with all Pillow versions.

Credit to https://github.com/tensorflow/models/issues/11040 for showing an easy solution.

```py
import fiftyone as fo
import fiftyone.zoo as foz

dataset = foz.load_zoo_dataset("quickstart-video", max_samples=2)

# previously failed with `Pillow==10` installed; now succeeds
dataset.draw_labels("/tmp/draw-labels", overwrite=True)
```

### DIFF
--- a/eta/core/annotations.py
+++ b/eta/core/annotations.py
@@ -1663,7 +1663,7 @@ def _draw_bounding_box(
 
     # Title background
     if title_str:
-        textw, texth = font.getsize(title_str)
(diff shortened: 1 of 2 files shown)


benjaminpkane (approved):
LGTM

PR #618, opened 2024-02-12
Azure extra cloud client kwargs

swheaton:
Supersedes previous PR related to this branch. I don't want to deal with `main` branch anymore.

Add in **kwargs so that the user can pass optional parameters to azi.DefaultAzureCredential object, in the vein of all the other StorageClient classes in this file.


benjaminpkane (approved):
Sounds good 👍

PR #628, opened 2024-07-09
Removes Python2 Support (which started as purging `six`)

findtopher:
I went in to purge `six` from the repo and found that we're supporting all kinds of very EOL Python...

This is my attempt to remove support for those versions.

This was accomplished with `pyupgrade --py3-plus`, `2to3 -wn -f dict -f buffer -f idioms -f set_literal -f ws_comma`, manually adding replacements for `future.utils`, and then manually cleaning up anything else that came from the pre-commit hooks.

There's no tests so I don't really have a way to validate these changes...


findtopher:
looks like it's created some circular references - will roll back to check this later

findtopher:
maybe I'll try again some day when I have time to resolve the import errors 🤷

PR #633, opened 2024-09-10
Upgrades to support remote zoo models

brimoor:
Supporting PR for https://github.com/voxel51/fiftyone/pull/4786

### DIFF
--- a/eta/core/models.py
+++ b/eta/core/models.py
@@ -1038,12 +1038,17 @@ class Model(Serializable):
 
     Attributes:
         base_name: the base name of the model (no version info)
-        base_filename: the base filename of the model (if any, no version info)
+        base_filename: the base filename or directory of the model (if any)
+            (no version info)
+        subdir: the model's subdirectory (if any)
         manager: the ModelManager instance that describes the remote storage
             location of the models_dir (if any)
-        version: the version of the model (if any)
+        author (optional): the author of the model
+        version: (optional) the model version
+        url (optional): the URL where the model is hosted
+        source (optional): the source of the model
+        license (optional): the license under which the model is distributed
         description: the description of the model (if any)
-        source: the source of the model (if any)
         size_bytes: the size of the model on disk (if any)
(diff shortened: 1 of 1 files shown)


eta/core/models.py
```diff
@@ -1472,70 +1496,103 @@ class ModelsManifest(Serializable):
...
             models: a list of Model instances
+            name (optional): a name for the manifest
+            url (optional): the source location of the manifest
         """
-        self.models = models or []
+        if models is None:
+            models = []
+
+        if name is not None:
+            subdir = os.path.join(*name.split("/"))
```
swheaton, line 1511: This seems like it is a no-op? You're splitting then joining it back together again. Do you mean to take out the file name? If so, why not use `os.path.dirname(name)`?

brimoor: It's for windows. Converting `org-name/model-name` to `org-name\model-name` if necessary

eta/core/models.py
```diff
@@ -1383,17 +1405,11 @@ def parse_name(name):
...
-
-        Raises:
-            ModelError: if the model name was invalid
         """
-        chunks = name.split("@")
+        chunks = name.rsplit("@", 1)
         if len(chunks) == 1:
             return name, None
 
-        if chunks[1] == "" or len(chunks) > 2:
```
swheaton, line 1394: What happens now then in the case that `version==""`? I guess it would be like `ab@`.
Another case of different behavior is `a@b@1.2`.

eta/core/models.py
swheaton: mm I see. A little confusing that `name` is assigned to `subdir` then.

eta/core/models.py
brimoor: The `@` feature is undocumented and unused in FO to-date, this versioning feature is mostly leftover cruft from 2017. It does work though!

swheaton: ok

swheaton (approved):
alright lgtm then

PR #637, opened 2024-09-18
it is time to deprecate python 3.8

findtopher:
## What changes are proposed in this pull request?

Deprecate python 3.8 as [as foretold by prophecy](https://docs.voxel51.com/deprecation.html#python-3-8)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,13 +29,7 @@ FROM $BASE_IMAGE
 # Install ETA
 #
 # Notes:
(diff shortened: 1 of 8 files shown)


brimoor (changes requested):
Can you also update the [README](https://github.com/voxel51/eta/blob/103631769553965f11167a4a4ba0a0025d3b2db5/README.md?plain=1#L25)?

brimoor (approved):
LGTM

PR #658, opened 2025-07-26
Updating signing creds logic to support identity pool + impersonation

prernadh:
Current logic for gcs signing credentials does not work for identity pool credentials type (WIF) and for impersonated credentials. This PR updates the logic so we can generate signed URLs for both.


brimoor (approved):
Static review LGTM ✅

PR #666, opened 2025-08-05
chore: remove Python 2 compatibility, black, pylint

findtopher:
I'm really not even sure who to have review this...

This is the simplest form of refactoring out Python 2 support, since we made python 3.9 the requirement some number of months ago.

I built this library and deployed topher.ephem.fiftyone.ai using it - as far as I can tell everything is working, but I'm not entirely sure how to test all the paths in this library.

Many of the changes in this PR are made by having run `pre-commit`, which clearly hadn't been run in quite some time.

If someone has a recommendation for making sure we know that this works, I'd love to hear it!


docs/core_dev_guide.md
findtopher, line 1: All changes in this file were made automatically by `pre-commit`

docs/custom_opencv_builds.md
findtopher, line 1: all changes in this file were made automatically by `pre-commit`

docs/export_tf_graphs_guide.md
findtopher, line 1: Changes in this file were made by `pre-commit`

topher updated the Copyright date

eta/core/cli.py
```diff
@@ -4156,7 +4142,7 @@ def _parse_google_drive_mime_type(mime_type):
 
 
 def _parse_datetime(datetime_or_str):
-    if isinstance(datetime_or_str, six.string_types):
```
afoley587, line 4159: `six.string_types` is a tuple:


```
>>> six.string_types
(<class 'str'>,)
>>> str
<class 'str'>
>>> 
```

brimoor (commented):
Gave this a quick glance. Nothing here scares me, but I have no good suggestions on how to actually test everything to verify.

Does `pre-commit` also offer a no bug guarantee by chance? 😉

eta/core/cli.py
findtopher: I got my replacement from the six [documentation](https://six.readthedocs.io/#six.string_types)

🤷

findtopher:
all the `six.string_types` replacements either fall _directly_ into an `isinstance()` call, or a `_parse_key` call, which indirectly uses them in a `isinstance()` call.

I feel pretty safe with this one...

```
def _parse_key(d, key, t, default):
    if key in d:
        val = d[key]
        if t is None or isinstance(val, t):
            # Return provided value
            return val, True
```

PR #667, opened 2025-08-14
Updating signing creds logic to not check for _is_default_credentials

prernadh:
Irrespective of whether they are default credentials or not, when we have identity pool or impersonated credentials we need to follow a different code logic.


prernadh:
@findtopher Apart from merging into develop, anything else you would need me to do?

PR #669, opened 2025-08-29
Add new training data key training_data key for Model class

Br2850:
Add new training_data key from ML to the core Model class so that subsequent SDK methods can recognize and import this new key, especially ZooModel


afoley587 (approved):
This looks good to me!

Our typical `git flow` would be to go to a `release/v0.15.1` path. But since this is the only ETA change between develop and main, that seems pedantic for this go-around.

All these changes seem good to me!

PR #675, opened 2025-11-07
chore(.github) AS-1099 dependabot, codeowners, templates

afoley587:
We utilize dependabot in other repos to keep our github actions up-to-date and somewhat in sync. This helps us get a consistent look/feel across our actions (as params change between versions).

This PR also adds CODEOWNERS and a PR template. These likely aren't going to be used too much, but CODEOWNERS with the dependabot things makes sure that the DB PRs are routed to aloha for review.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,9 @@
 * @voxel51/developers
 
 # Aloha!
-.github/  @voxel51/aloha-shirts
-setup.py  @voxel51/aloha-shirts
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
mo-getter: Do we also want to code-own any other files (possibly the following)?

- pyproject.toml
- RELEASING.md
- Dockerfile

mo-getter (approved):

PR #718, opened 2026-07-24
Replace pysftp with native paramiko SFTP client

benjaminpkane:
pysftp is unmaintained (last release 2016) and does `from paramiko import DSSKey` at module import, which paramiko >=4 removed — this is what forced the `paramiko<4` pin (FOEPD-1886) and blocks CVE-2026-44405 remediation downstream.

`SFTPStorageClient` now drives `paramiko.SFTPClient` directly:
- `put`/`putfo`/`get`/`getfo`/`remove`/`rmdir` map 1:1
- small helpers reproduce pysftp's recursive `put_r`/`get_r`/`makedirs`, including its octal-as-decimal `mode` convention
- host-key verification behavior preserved (`load_system_host_keys` + reject unknown, matching pysftp's default `CnOpts`)

Lifts the pin to `paramiko>=3,<6` and drops the `pysftp` extra. Works against paramiko 3 (current resolve) and 5.

Note: the repo has no SFTP test coverage, so this was validated by API-surface parity; the storage extra's consumers (downstream `SFTPStorageClient` users) should smoke-test before a release is cut.

### DIFF
--- a/eta/core/storage.py
+++ b/eta/core/storage.py
@@ -8,7 +8,7 @@ This module currently provides clients for the following storage resources:
 - S3 buckets via the `boto3` package
 - Google Cloud buckets via the `google.cloud.storage` package
 - Google Drive via the `googleapiclient` package
-- Remote servers via the `pysftp` package
+- Remote servers via the `paramiko` package
 - Web storage via HTTP requests
 - Local disk storage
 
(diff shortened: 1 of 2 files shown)


kevin-dimichel (approved):
Aloha changes look fine (as long as the storage devs approve).

PR #738, opened 2026-09-18
fix(foepd-4748): authorize HEAD when signing Azure Blob Storage URLs

swheaton:
# Rationale

Signing a URL for an HTTP HEAD raised `KeyError: 'HEAD'` on Azure Blob Storage. `AzureStorageClient._permissions` mapped only GET, PUT and DELETE to a SAS permission, and `generate_signed_url()` indexed that dict directly, so the failure happened before any request went out — deterministic, and independent of the blob, the container or the credentials.

A caller that needs only an object's size signs a HEAD and reads `Content-Length` rather than downloading the object. That call path is unusable on Azure today: every attempt raises. FiftyOne Enterprise hits it through `fiftyone.core.storage.get_file_size()`, where it fails 100% of one feature's work for any customer whose media lives in Azure (FOEPD-4748). The S3 and MinIO clients derive a botocore method from the verb, and GCS forwards the verb into the signature, so those clients already signed a HEAD.

Only the account-key and user-delegation-key credential modes reach the permission lookup. A client configured with a SAS token appends that token instead of signing, so it never saw the error.

## Changes

- Map `HEAD` to `BlobSasPermissions(read=True)`. Read is what authorizes `Get Blob Properties`, the REST operation a HEAD on a blob maps to, and it is the narrowest permission Azure offers — there is no properties-only letter.
- Raise a `ValueError` naming the verb, and the verbs that are supported, for a method the client cannot sign, so the next unsupported method is not a bare `KeyError` either.
- Document `HEAD` in the `method` argument of all three clients that sign URLs, and the new exception in `Raises:`.
- Add `tests/test_storage_signed_urls.py`.
- Install the `storage` extra in the `test` CI job, which the new tests need.

## Testing

`pytest tests/ -v` passes: 52 tests, of which 20 are new.

The tests fabricate credentials and never reach a provider. Constructing a client and signing a URL are both offline, so they exercise the real signing code — Azure's HMAC over the account key, botocore's presigner, and GCS's local RSA signature — with no network and no account.

Each provider gets two kinds of assertion, because they answer different questions:

- That a signed URL comes back at all, with a signature parameter, for GET and HEAD. On Azure this also asserts the SAS permission letter for every method in the map (`r`, `c`, `d`), since this change edits that map.
- That the verb actually reaches the signer. A signature parameter is present however the verb is handled, so on its own the first assertion would pass for a client that signed every method as a GET.

Verified by mutation — each of these fails the suite:

| Mutation | Tests failing |
| --- | --- |
| Drop the `HEAD` entry from `_permissions` | 3 |
| `HEAD` mapped to `delete=True` instead of `read=True` | 3 |
| Restore the bare `KeyError` lookup | 1 |
| Azure `PUT` mapped to `read=True` | 1 |
| boto signs `get_object` regardless of the verb | 2 |
| GCS signs `GET` regardless of the verb | 1 |

The ffmpeg CI jobs do not install the `storage` extra, so the file probes for the cloud SDKs and skips when they are absent. It probes the SDKs rather than catching an `ImportError` from `eta.core.storage`: catching one there would skip the whole file, and report a missing extra, for a genuine broken import in the module under test. Verified both ways — with the SDKs hidden the file skips; with an undefined symbol imported in `storage.py` it errors rather than skipping.

## Notes

`read` authorizes more than a HEAD strictly needs: the same signed URL also permits a full `Get Blob` download. Azure has no narrower permission, so a signed HEAD URL is as powerful as a signed GET URL by construction.

Unrelated and left alone: `origin/main` currently fails the repo's pinned `black` hook on two pre-existing spots in `eta/core/storage.py` (a line-continuation import at the top of the file, and an 82-character line four lines from this change). No CI job runs `black`, and reformatting them here would be churn in a review.


swheaton:
done elsewhere, don't need this here

